PR #77, opened 2026-01-12
[ADR 008] nix, pulumi, image workflow restructuring

### LINKED ISSUE #75: Confusing separation of image build and push jobs in nix.yml workflow
## Problem

The nix.yml reusable workflow separates image building and image pushing into separate jobs, which creates confusion and potentially wastes CI minutes.

## Current Behavior

When using the nix.yml workflow:
1. Image packages (ending in `-image`) build successfully ✓
2. If ANY check fails (treefmt, pre-commit, etc.), the push job is silently skipped
3. The workflow shows "success" for the build jobs, making it unclear why images aren't in the registry
4. This wastes CI minutes building images that never get pushed

## Example

In https://github.com/cavinsresearch/zeus/actions/runs/20698896732:
- `packages.x86_64-linux.klotho-image` ✓ built successfully
- `packages.x86_64-linux.poseidon-image` ✓ built successfully  
- `checks.x86_64-linux.treefmt` ✗ failed (formatting issue)
- `🐳 push ${{ matrix.package-name }}` - skipped (no indication why)

Result: We thought the images were pushed, but they weren't in GHCR.

## Suggested Improvement

One of:
1. **Combine build+push**: Make image packages build and push in a single job (like docker-build.yml does)
2. **Clearer feedback**: Make it very obvious when push is skipped and why
3. **Fail fast**: Don't build images if checks are failing (to save CI minutes)

## Context

The commit message in 37ebf1b5f indicates auto-detection and pushing should work: "The upstream nix.yml workflow now auto-detects and pushes images (any package ending in -image)". However, the conditional push logic makes this behavior unclear to users.

jmmaloney4:
Resolves #75 by decoupling image building from check execution:

- Modified build job to exclude images (is_image: true) via matrix exclude
- Renamed push-images to build-and-push-images and removed build dependency
- Images now build and push independently, even if checks fail
- This prevents the confusing scenario where images build successfully but silently skip push when checks fail

Also adds ADR-008 documenting options for SHA-tagged image deployment coordination between nix.yml and pulumi.yml workflows. The ADR evaluates six approaches for passing image tags from build to deploy, recommending a hybrid of convention-based tagging with optional orchestration.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -73,6 +73,9 @@ jobs:
       max-parallel: 1
       matrix:
         include: ${{ fromJson(needs.detect.outputs.matrix_include) }}
+        exclude:
+          # Exclude images from this job - they're built and pushed separately
+          - is_image: true
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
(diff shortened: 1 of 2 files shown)


jmmaloney4:
Superseded by #196, which keeps only the workflow change that decouples image publishing from the general build matrix. The ADR-heavy scope was intentionally split out.

Issue #97, opened 2026-02-21
compute-flake-build-matrix: detect job reports no work when only formatter is discovered


jmmaloney4:
## Problem
The `❄️ nix` reusable workflow's `☃️ detect flake outputs` job can incorrectly produce:
- `matrix_include = []`
- `has_work = false`

…even when the target flake has buildable `packages`/`checks` outputs.

## Evidence
From a failing run:

- `All detected outputs: [{"attr":"formatter","category":"formatter","system":"unknown","name":"default","flake_attr":".#formatter",...}]`
- `Computed include (uncached only): []`
- `Has work: false`

So detection only saw `formatter`, then filtered it out, resulting in "no work to do".

### 1) Selector rewrites output shape in a way that breaks downstream assumptions
` .github/actions/compute-flake-build-matrix/select.nix ` currently rewrites each top-level output to `outputs.<category>.<system>` (or `{}`).

That means the attr paths emitted by `nix-eval-jobs` may no longer have the expected `category.system.name` structure. In some cases only direct derivations (like `formatter`) are emitted.

### 2) Parser assumes attr path is always `category.system.name`
In ` .github/actions/compute-flake-build-matrix/main.sh `, jq parsing assumes:
- `category = split(".")[0]`
- `system = split(".")[1]`
- `name = split(".")[2]`
- `flake_attr = .# + attr`

If attrs are `category.name` after selection, `system`/`name` are miscomputed and `flake_attr` becomes invalid for build matrix use.

### 3) Category filter drops formatter entirely
Later filter only keeps `packages|checks` (plus `is_image` special-case), so if only `formatter` is discovered, include becomes empty by design.

### Option A (recommended): narrow selector + robust attr parsing
1. In `select.nix`, return only selected-system `packages` and `checks`, and set `recurseForDerivations = true` so nix-eval-jobs traverses those attrsets.
2. In `main.sh`, parse both attr shapes:
   - `packages.<system>.<name>` (full path)
   - `packages.<name>` (selected-system path)
   and reconstruct `flake_attr` with current system when needed.

Suggested `select.nix` pattern:

```nix
system: outputs:
let
  pick = name:
    if builtins.hasAttr name outputs
      && builtins.isAttrs (builtins.getAttr name outputs)
      && builtins.hasAttr system (builtins.getAttr name outputs)
    then (builtins.getAttr system (builtins.getAttr name outputs)) // { recurseForDerivations = true; }
    else {};
in {
  packages = pick "packages";
[...]

PR #98, opened 2026-02-21
fix(compute-flake-build-matrix): narrow select.nix to packages/checks and handle 2-part attrs

jmmaloney4:
## Summary

- Fix issue #97 where `compute-flake-build-matrix` detected no work when only formatter was in flake outputs
- Narrowed `select.nix` to only `packages` and `checks` categories (instead of iterating all output categories)
- Updated `main.sh` jq parsing to handle both 2-part attrs (`packages.foo`) and 3-part attrs (`packages.x86_64-linux.foo`), reconstructing the correct `flake_attr` for the 2-part case

### DIFF
--- a/.github/actions/compute-flake-build-matrix/select.nix
+++ b/.github/actions/compute-flake-build-matrix/select.nix
@@ -8,7 +8,7 @@ let
 
   pick = name:
     let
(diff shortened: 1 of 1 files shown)


jmmaloney4:
@codex review

PR #236, opened 2026-05-31
fix(monitor): set D1 readReplication explicitly

### LINKED ISSUE #235: bug(monitor): D1 update can fail with null read_replication on existing uptime DBs
## Bug description
Updating a stack that uses `UptimeMonitor` can fail on the `cloudflare:index:D1Database` resource even when the component only created the D1 database with `accountId` and `name`.

Observed failure from `garden`:

```text
error: [ERROR] PUT "https://api.cloudflare.com/client/v4/accounts/622586854078fb4b8547f4d2289c15ab/d1/database/41bfcd5b-0d90-4e77-9602-5b2705cda7d1": 400 Bad Request {"success":false,"errors":[{"code":7400,"message":"Invalid property: read_replication => Expected object, received null"}]}
```

## Impact
This breaks `pulumi up` for stacks consuming `@jmmaloney4/sector7/monitor` when the D1 database already exists and Pulumi decides the DB needs an update.

## What I verified
Repo: `jmmaloney4/garden`
Stack path: `deploy/services/uptime`
sector7 version: `0.11.0`
@pulumi/cloudflare version: `6.17.0`

`deploy/services/uptime/index.ts` creates the component like this:

```ts
const monitor = new UptimeMonitor(servicePrefix, {
  accountId,
  name: servicePrefix,
  monitors,
  webhookUrl,
  cronSchedule,
  apiToken,
});
```

The published `UptimeMonitor` implementation creates the D1 resource with only:

```ts
this.d1Database = new cloudflare.D1Database(`${name}-d1`, {
  accountId: args.accountId,
  name: dbName,
}, resourceOpts);
```

So sector7 is not explicitly setting `readReplication`.

## State/provider evidence
The stack state for the existing D1 resource contains:

```json
"readReplication": {
  "mode": "disabled"
}
```

And the installed provider code includes `readReplication` in update inputs:

```js
resourceInputs["readReplication"] = state?.readReplication;
```

(from `@pulumi/cloudflare@6.17.0` `d1database.js`)

This suggests the failure path is:

1. `UptimeMonitor` creates or adopts a D1 DB without setting `readReplication`
2. refresh/imported state includes `readReplication: { mode: "disabled" }`
3. provider update serializes `read_replication` incorrectly as `null`
4. Cloudflare rejects the PUT

## Reproduction
From `garden`:

```bash
pulumi -C deploy/services/uptime up -s dev --refresh
```

Then approve the update. The D1 update fails with the 400 above.

## Expected behavior
One of these should be true:

- `UptimeMonitor` should be resilient to this provider/API mismatch for default D1 databases, or
- sector7 should document/pin/work around the bad provider behavior, or
- if there is a safe explicit setting for `readReplication`, sector7 should set it to avoid the null serialization path.

## Notes
I do not yet know whether the final fix belongs in sector7 or only in the upstream Pulumi Cloudflare provider. But sector7 consumers hit it through `UptimeMonitor`, so this repo is a reasonable place to track mitigation/workaround guidance.

jmmaloney4:
## Summary
- explicitly set `readReplication: { mode: "disabled" }` when `UptimeMonitor` creates a D1 database
- add a unit test asserting the D1 resource carries the explicit read-replication object
- document the workaround in the PR rationale via issue link

## Why
`@pulumi/cloudflare` 6.17.0 can fail D1 updates after refresh with:

```text
Invalid property: read_replication => Expected object, received null
```

`UptimeMonitor` currently relies on the provider default. This PR makes the default explicit so the created D1 resource state matches the observed Cloudflare-side default (`{ mode: "disabled" }`) and avoids the null-serialization path we hit in garden.

## Validation
- `pnpm --filter @jmmaloney4/sector7 test -- --run packages/sector7/tests/uptime-monitor.test.ts`
- `pnpm --filter @jmmaloney4/sector7 build`

Closes #235

---

> [!NOTE]
> **Low Risk**
> Small infrastructure-as-code default change with no runtime monitor behavior change; only affects newly created D1 resources via this component.
> 
> **Overview**
> When `UptimeMonitor` provisions a new D1 database, it now sets **`readReplication: { mode: "disabled" }`** on the `cloudflare.D1Database` resource instead of leaving that field implicit. That pins the provider input to the same shape Cloudflare already uses by default, so **`pulumi refresh` / subsequent updates** with `@pulumi/cloudflare` ≥ 6.17.0 do not send `read_replication: null` and trip the API validation error.
> 
> A unit test asserts the mocked D1 resource inputs include that explicit read-replication object.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f61249218696d7f3f90d63f7f15e81b8e9282e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sector7/monitor/uptime-monitor.ts
+++ b/packages/sector7/monitor/uptime-monitor.ts
@@ -272,6 +272,8 @@ export class UptimeMonitor extends pulumi.ComponentResource {
 					{
 						accountId: args.accountId,
 						name: dbName,
+						// Explicitly set to avoid the @pulumi/cloudflare >= 6.17.0
+						// null-serialization bug on D1 updates after refresh.
 						readReplication: { mode: "disabled" },
 					},
 					resourceOpts,
--- a/packages/sector7/monitor/uptime-monitor.ts
+++ b/packages/sector7/monitor/uptime-monitor.ts
@@ -272,6 +272,7 @@ export class UptimeMonitor extends pulumi.ComponentResource {
 					{
 						accountId: args.accountId,
 						name: dbName,
+						readReplication: { mode: "disabled" },
 					},
 					resourceOpts,
 				);
--- a/packages/sector7/tests/uptime-monitor.test.ts
+++ b/packages/sector7/tests/uptime-monitor.test.ts
@@ -79,7 +79,9 @@ describe("UptimeMonitor", () => {
 		await resolveOutput(monitor.worker.id);
 		await resolveOutput(monitor.cronTrigger.id);
 
-		expect(findResource("basic-d1")).toBeDefined();
+		const d1 = findResource("basic-d1");
+		expect(d1).toBeDefined();
+		expect(d1?.inputs.readReplication).toEqual({ mode: "disabled" });
 		expect(findResource("basic-kv")).toBeDefined();
 		expect(findResource("basic-worker")).toBeDefined();
 		expect(findResource("basic-cron")).toBeDefined();


jmmaloney4:
Addressed the Gemini review note by adding an inline comment above the explicit `readReplication: { mode: "disabled" }` setting, then re-ran the uptime monitor test file and package build locally.

PR #245, opened 2026-06-03
feat(nix-image): coordinate push phase to dedupe shared base layers (ADR-029)

jmmaloney4:
## Summary

Images built from one monorepo share a large nix base layer but differ only in a small top layer. Pulumi pushes independent `NixImage`s **concurrently**, so every push races to upload the same base blob — the per-blob existence check (`HEAD /v2/<name>/blobs/<digest>`) returns 404 for all of them before any finishes, and the identical multi-hundred-MB base is uploaded once per image. Observed in yard's cron stack, where `discovery-runner`, `dbt`, and `pipeline-runner` each re-push the same `sha256:…` base on every rebuild.

This adds a **push-coordination** layer that serializes the push phase while keeping the `nix build` phase parallel.

Implements **ADR-029** (`docs/internal/designs/029-nix-image-push-coordination.md`).

## Design

**`NixImagePushGroup`** — a tiny in-memory coordinator (no Pulumi state). `NixImage` calls `group.dependencies()` before constructing its push `Command` (feeding the result into `dependsOn`) and `group.register(cmd)` after.

- `"serial"` (default): each push waits for the previous one — correct regardless of which layers are shared.
- `"primer"`: only the first push runs alone; later pushes wait for it then run concurrently (optimal when all images share a base — the primer uploads it once, the rest skip it and upload unique layers in parallel).

**`pushGroup?: NixImagePushGroup | false`** on `NixImageArgs`:

- **omitted (default):** the image joins an internal group **keyed on `artifactRegistryUrl`**. Same-registry pushes serialize (shared layer uploaded once, the rest skip via `HEAD`); different-registry pushes stay parallel. **Zero consumer changes** — the "just works" path.
- **a `NixImagePushGroup`:** join an explicit group to widen/narrow the set or opt into `"primer"`.
- **`false`:** opt out; push runs unordered (pre-PR behavior).

### Why this beats a consumer-side `dependsOn`

Only the push `Command` gets the added `dependsOn`, so **builds stay parallel**. A consumer-level `dependsOn` between whole `NixImage`s (the current yard workaround) also serializes the `nix build`s. This lets those consumers delete the workaround.

### Dedup boundary (why key on registry)

Layer blobs dedupe **within a registry** — for GCP Artifact Registry, within one repository resource. A layer present anywhere in that repo satisfies the `HEAD` check, so a later same-registry push skips it. Across different registries nothing is shared (skopeo doesn't implement the OCI cross-repo `mount` API). So coordination only needs to group pushes destined for the same registry.

## Notable implementation detail

The coordinator is deliberately kept out of the args passed to `super(...)`. Once `register()` stores a push `Command`, the group transitively references this component's own child → leaking it into the registered inputs creates a registration cycle (manifested as a hang). Destructuring `pushGroup` out before `super` avoids it. Regression-covered by the build-mode integration test.

## Test plan

- [x] `tsc -p packages/sector7/tsconfig.build.json --noEmit` — clean.
- [x] `vitest run` — **173 passed** (4 new pure unit tests for `NixImagePushGroup` serial/primer/empty; 3 new `NixImage` integration tests: build-mode registration, resolve-mode non-registration, `false` opt-out).
- [x] `just fmt` + `just lint` — clean.

## Follow-ups (not in this PR)

- Minor version bump + release (`just cut minor`) so consumers can pick it up.
- yard: once released, replace the `dependsOn` workaround (addendalabs/yard#1563) with the default coordination.

---

> [!NOTE]
> **Medium Risk**
> Default behavior serializes same-registry pushes (deploy timing change) but only affects performance, not correctness; no auth or push-script changes.
> 
> **Overview**
> Adds **push coordination** to `NixImage` so concurrent registry pushes no longer race to re-upload the same shared nix base layer. A new in-memory **`NixImagePushGroup`** (`serial` or `primer`) wires `dependsOn` only on push `Command`s, leaving **`nix build` parallel**.
> 
> By default, images sharing the same **`artifactRegistryUrl`** join an internal serial group (zero config); callers can pass an explicit group, or **`pushGroup: false`** for the old unordered behavior. **`resolve`** mode never coordinates. **`pushGroup`** is omitted from component inputs passed to `super()` to avoid a registration cycle.
> 
> **ADR-029** documents the design. **`barrel-guard-gateway.ts`** is tightened to a type-only guard that fails the build if `gateway` is re-exported from the main barrel. New unit and integration tests cover group strategies and `NixImage` wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6cc182ffc433b3dc459e7a6c3c46df740115f158. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/internal/designs/029-nix-image-push-coordination.md
+++ b/docs/internal/designs/029-nix-image-push-coordination.md
@@ -1,0 +1,229 @@
+---
+id: ADR-029
+title: Coordinated Push Phase for NixImage (Shared Base Layer Deduplication)
+status: Proposed
+date: 2026-06-03
+deciders: [jmmaloney4]
+consulted: []
+tags: [design, adr, nix-image]
+supersedes: []
+superseded_by: []
+links: [ADR-017]
+---
+
+# Context
+
+Repositories that build several container images from one monorepo (yard, zeus)
+almost always produce images that **share a large base layer** and differ only
+in a small top layer. With nix2container this is the normal outcome: the shared
+runtime closure is one content-addressed store path, so every image references
+the *same* layer digest.
+
+`NixImage` (ADR-017) builds each image with a `NixOutput` (the `nix build`) and
+pushes it with a `command.local.Command` running `skopeo copy nix:<store-path> docker://<ref>`. Pulumi has no dependency edge between independent `NixImage`s,
+so it runs their pushes **concurrently**. Each push independently probes the
+registry for each layer (`HEAD /v2/<name>/blobs/<digest>`) and uploads the ones
+that are missing.
+
+When N images that share a base layer are pushed at once, all N `HEAD`s for the
+shared blob return 404 before any push has finished uploading it, so the
+identical multi-hundred-MB base is uploaded N times. Only one copy is kept; the
+other N-1 uploads are wasted bandwidth and wall-clock. This was observed in
+yard's cron stack, where three images (`discovery-runner`, `dbt`,
+`pipeline-runner`) each re-push the same `sha256:…` base on every rebuild.
+
+**Deduplication boundary.** Registries deduplicate layer blobs per *registry*,
+not globally. For GCP Artifact Registry the boundary is the **repository
+resource** (`PROJECT/REPO`): images pushed under different image names within
+one AR repository share stored blobs, and a layer already present anywhere in
+that repository satisfies the `HEAD` existence check, so a later push skips it.
+Across *different* AR repositories nothing is shared (skopeo does not implement
+the OCI cross-repo blob `mount` API — only crane/gcrane do). The fix therefore
+only needs to coordinate pushes **destined for the same registry**.
+
+In scope: avoiding concurrent re-upload of shared layers from `NixImage`'s push
+phase, with a default that requires no consumer changes. Out of scope:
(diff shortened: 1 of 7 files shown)


jmmaloney4:
### Review gate status (maintainer-approved waiver)

- **Cursor Bugbot:** reviewed and passed (Medium Risk — the by-design default behavior change; no correctness issues). Re-runs on the latest commit.
- **Gemini Code Assist:** hit its **daily quota** ("wait up to 24 hours") and cannot review today. Per maintainer (jmmaloney4) decision, the Gemini review gate is **explicitly waived** for this PR rather than blocking ~24h.
- **Copilot:** not enabled on this repo.

### Validation (CI substitute)

`nix flake check -L` is **green**: `tsc`, `vitest` (173 passed), `treefmt`, `pre-commit`, `renovate-config` all pass.

Note: commit `2c6c413` also repairs a **pre-existing** failure of `checks.tsc` (a stale gateway barrel-guard, red on `main` before this branch) so the flake check passes on this PR. The `add-to-project` CI job fails for an unrelated infra reason (`nix: command not found` on its runner).

PR #249, opened 2026-06-08
chore(nix-image): source skopeo from jackpkgs, auto-bump pin via Renovate

jmmaloney4:
## Summary

The nix-image push scripts (`nix-image-push.sh`, `nix-image-build-push.sh`) fetched `skopeo-nix2container` directly from `github:nlewo/nix2container/76be9608…#skopeo-nix2container`. This points them at jackpkgs' re-exported package instead — `github:jmmaloney4/jackpkgs/791f4529…#skopeo-nix2container` (see [jackpkgs#296](https://github.com/jmmaloney4/jackpkgs/pull/296)) — so the skopeo source is centralized in one place.

jackpkgs currently pins `nix2container` at the **same** `76be9608` commit, so this resolves to the **identical skopeo store path today**. It's a consolidation, not a behavior/version change.

### Why this matters
- **Single source of truth.** When nix2container needs bumping, it happens once in jackpkgs rather than being hardcoded across consumers.
- **Reproducible + auto-maintained.** The pin stays a specific commit (reproducible), and a new Renovate manager bumps it automatically.

## Changes
- **`packages/sector7/scripts/nix-image-push.sh`** + **`nix-image-build-push.sh`** — 4 skopeo refs repointed to `jmmaloney4/jackpkgs`.
- **`renovate/nix.json`**:
  - New custom manager: matches `nix run github:owner/repo/<40-hex>#…` commit pins in shell scripts and bumps them to the latest commit on `main` via the `git-refs` datasource.
  - Tightened the existing version-tag manager with a `#` boundary so it no longer false-matches the leading digits of a commit SHA (previously `…/76be9608#…` matched as version "76").
- **`packages/sector7/tests/renovate-nix-regex.test.ts`** — 4 new tests (match + mutual exclusion for both managers).
- **`renovate/README.md`** — documented the shell-script flake-pin managers.

## Test plan
- `vitest run` — 178 passed (17 files), incl. 4 new regex tests.
- `tsc --noEmit` — clean.
- `renovate-config-validator --strict` — passes (`.github/renovate.json`, `.github/renovate.json5`, `renovate/nix.json`).
- `bash -n` on both scripts — clean.
- `pnpm run build` — tsc + shell-script copy succeed; `dist/` scripts carry the jackpkgs ref.

## Risks
- Low. Same skopeo store path as before today. The only new moving part is the Renovate `git-refs` manager, which has been validated and is covered by tests.

## Follow-up
- Release **v0.13.0** after merge, then bump zeus (`package.json` + `atlas/package.json`) from `v0.11.0` → `v0.13.0` to pick this up.

### DIFF
--- a/packages/sector7/scripts/nix-image-build-push.sh
+++ b/packages/sector7/scripts/nix-image-build-push.sh
@@ -98,7 +98,7 @@ chmod 600 "${AUTH_FILE}"
 if [ "${SCRIPT_MODE}" = "resolve" ]; then
   # Resolve-only: inspect the already-pushed image to get its digest
   echo "--- resolving digest ---"
-  nix run github:nlewo/nix2container/76be9608a7f4d6c985d28b0e7be903ae2547df3e#skopeo-nix2container -- \
+  nix run github:jmmaloney4/jackpkgs/791f4529199a9c13a8a66f5ddd2eb642198447c5#skopeo-nix2container -- \
     --insecure-policy inspect --format '{{.Digest}}' \
     --authfile "${AUTH_FILE}" \
     docker://"${FULL_TAG}" |
@@ -115,7 +115,7 @@ else
 
   # Push the image
   echo "--- skopeo copy ---"
-  nix run github:nlewo/nix2container/76be9608a7f4d6c985d28b0e7be903ae2547df3e#skopeo-nix2container -- \
+  nix run github:jmmaloney4/jackpkgs/791f4529199a9c13a8a66f5ddd2eb642198447c5#skopeo-nix2container -- \
     --insecure-policy copy --digestfile "${DIGEST_FILE}" \
     --authfile "${AUTH_FILE}" \
     "${IMAGE_PATH}" \
--- a/packages/sector7/scripts/nix-image-push.sh
+++ b/packages/sector7/scripts/nix-image-push.sh
@@ -85,7 +85,7 @@ chmod 600 "${AUTH_FILE}"
 
 if [ "${SCRIPT_MODE}" = "resolve" ]; then
   echo "--- resolving digest ---"
-  nix run github:nlewo/nix2container/76be9608a7f4d6c985d28b0e7be903ae2547df3e#skopeo-nix2container -- \
+  nix run github:jmmaloney4/jackpkgs/791f4529199a9c13a8a66f5ddd2eb642198447c5#skopeo-nix2container -- \
     --insecure-policy inspect --format '{{.Digest}}' \
     --authfile "${AUTH_FILE}" \
     docker://"${FULL_TAG}" |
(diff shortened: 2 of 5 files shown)


jmmaloney4:
Thanks @gemini-code-assist — both addressed in 8450b1f:

1. **`github-commits` datasource** (commit-SHA manager): adopted. Switched off `git-refs` + clone URL. Note: `renovate-config-validator --strict` still **requires** a `currentValue`/`currentValueTemplate` even for `github-commits`, so the default branch can't be left implicit — kept `currentValueTemplate: "main"` (jackpkgs' default branch).

2. **Version-tag boundary**: the intent is adopted, but the exact `(?=[\s#]|$)` lookahead can't be used — Renovate's regex managers run on **RE2**, which rejects lookahead/lookbehind assertions. Used an RE2-safe *consumed* terminator instead: `(?:[#\s]|$)`. This still excludes commit SHAs (digits followed by hex letters fail the class) **and** now matches version pins with no `#attribute`, which was your underlying concern. Added a test for the no-attr case.

`renovate-config-validator --strict` passes; regex tests green (179 total).

jmmaloney4:
Fixed in $(latest) — README now says `github-commits` to match the implementation. Thanks @gemini-code-assist.

/gemini review

jmmaloney4:
Addressed in 9c5b6c6:

- **Pre-release version tags**: added — `currentValue` now accepts an optional `-[a-zA-Z0-9.-]+` suffix (e.g. `v1.0.0-beta.1`), with a test.
- **Nix options + arguments**: replaced the flag-only `(?:-\w+ )*` with a lazy, RE2-safe `(?:\S+ )*?` that skips *any* options **and their arguments** between `nix` and `run`. Your `(?:\s+[^-\s]\S*)?` arg-matcher only captures one arg per flag and starts to mis-parse multi-arg options; the lazy skip handles the general case (e.g. `nix --extra-experimental-features flakes run github:...`, where `flakes` is an argument, not a flag). Avoided lookahead since Renovate runs on RE2. Added a test for that invocation.

Commit-SHA exclusion still holds (a 40-hex digest fails the version terminator). `renovate-config-validator --strict` passes; 181 tests green.

/gemini review

jmmaloney4:
Applied in $(git rev-parse --short HEAD) — both managers now use `\s+` / `(?:\S+\s+)*?` to tolerate arbitrary spacing and backslash-newline continuations, with a test. `renovate-config-validator --strict` passes; 182 tests green.

/gemini review

jmmaloney4:
@gemini-code-assist — declining the query-parameter (`?dir=…`) suggestions as **non-blocking / YAGNI**:

- These managers exist to track `nix run github:owner/repo/<ref>#attr` flake pins. The only such pin in this repo is `jmmaloney4/jackpkgs#skopeo-nix2container`, a top-level flake output — it has no `?dir=` subdir query, and there's no foreseeable reason it would.
- Over the last 4 rounds the regexes already gained: `github-commits` datasource, RE2-safe SHA exclusion, pre-release tags, option+argument skipping, and `\s+`/line-continuation tolerance. That covers the real-world surface.
- Adding query-string parsing trades real complexity for a case that can't occur here. If a consumer ever needs it, it's a one-line terminator tweak at that point.

All substantive feedback is addressed; remaining suggestions are low-priority hypotheticals. Considering review converged from my side.

Issue #165, opened 2026-05-04
GitHub App auth for add-to-project workflow


jmmaloney4:
## Context

PR #163 made `github-token` a required secret on the add-to-project reusable workflow. Now we need to set up the actual auth mechanism.

The plan is in `docs/internal/plans/2026-05-02-github-app-add-to-project.md`.

## Prerequisite: rename repo first

Before implementing this, we are renaming the repo from `toolbox` to `sector7` (see PR #164 for the rename plan). All naming should be centralized around `sector7`:

- App name: `sector7-project-bot` (not `toolbox-project-bot`)
- Secret names: `SECTOR7_PROJECT_BOT_APP_ID`, `SECTOR7_PROJECT_BOT_PRIVATE_KEY`
- All references in docs and workflows updated accordingly

## Steps after rename

1. Register GitHub App `sector7-project-bot` under ergodicsystems org
   - Permissions: Organization → Projects: Read and write
   - No webhooks
   - Install on ergodicsystems org
2. Store `SECTOR7_PROJECT_BOT_APP_ID` and `SECTOR7_PROJECT_BOT_PRIVATE_KEY` as org secrets on ergodicsystems (or repo secrets on callers)
3. Update the reusable workflow to use `actions/create-github-app-token@v2` internally — callers just pass `app-id` and `private-key` secrets
4. Update all caller workflows (garden, etc.) to pass the two secrets
5. Verify end-to-end with a test issue and PR

## Why GitHub App over PAT

- Scoped to exactly Projects read/write
- No expiry (private keys rotated on your schedule, not GitHub's)
- Not tied to a personal account

jmmaloney4:
## Companion fix for step 3: `create-github-app-token` needs `owner: ergodicsystems`

Reviewing the auth approach surfaced a gap in the ADR-022 target-state workflow: the `actions/create-github-app-token` step omits `owner`. With no `owner`, the action mints an installation token for **the caller repo's account**. That works for in-org callers (e.g. `ergodicsystems/hq`), but every caller outside the org — `jmmaloney4/garden`, `jmmaloney4/jackpkgs`, `cavinsresearch/zeus`, etc. — fails with *installation not found*, because the app is only installed on `ergodicsystems`.

**Fix:** pass `owner: ergodicsystems` (the account that owns both the app installation and project #1) so the token is always minted against the ergodicsystems installation regardless of which repo triggers the workflow.

[...]

PR #254, opened 2026-06-09
feat(ci): mint GitHub App token in add-to-project workflow

jmmaloney4:
## Summary

Implements **step 3** of the GitHub App auth plan (`docs/internal/plans/2026-05-02-github-app-add-to-project.md`, ADR-022) and the **`owner` companion fix** flagged in #165.

The reusable `add-to-project.yml` workflow now mints a GitHub App installation token internally via `actions/create-github-app-token@v3.2.0`. Callers pass `app-id` + `private-key` instead of a user PAT. Changes:

- **New `owner` input** (default `ergodicsystems`) passed to `create-github-app-token`. This is the companion fix: without it the action mints a token for the *caller repo's account*, so cross-org callers (`jmmaloney4/garden`, `cavinsresearch/zeus`, …) fail with *installation not found*. With it, the token is always minted against the account that owns the app installation, regardless of which repo triggers the run.
- **New `app-id` / `private-key` secrets** (optional, preferred auth).
- **`github-token` demoted** to an optional legacy fallback (`required: false`). This alone stops the org-wide "Secret github-token is required, but not provided" validation failures once callers stop passing it.
- **Guard step** fails with a clear error when neither app creds nor a fallback token are provided.
- Secrets mapped to job `env` so they can be used in step-level `if` (the `secrets` context isn't available in conditionals); the private key is referenced directly in `with:` and not exposed as a job env var.
- **ADR-022 updated**: target-state YAML synced to the real workflow (it had drifted and omitted `owner`); registration guidance set to **"Only on this account"** with the `owner` rationale.

## Test plan

- [x] `actionlint` passes clean on the workflow.
- [ ] End-to-end (blocked on plan steps 1–2, which are manual): register `sector7-project-bot` on ergodicsystems (Only on this account), store `SECTOR7_PROJECT_BOT_APP_ID` / `SECTOR7_PROJECT_BOT_PRIVATE_KEY`, then open a test issue/PR and confirm it lands on ergodicsystems/projects/1.

## Risks

- App auth is untested until the app exists and secrets are wired; the `github-token` fallback path preserves prior behavior in the meantime.
- `create-github-app-token` bumped to v3.2.0 (ADR mentioned v2); `owner` works identically.

## Remaining work (not in this PR — tracked in #165)

1. Register the GitHub App on ergodicsystems (**Only on this account**).
2. Store `SECTOR7_PROJECT_BOT_APP_ID` / `_PRIVATE_KEY` as ergodicsystems org secret + cavinsresearch org secret + repo secrets on each jmmaloney4 caller.
3. Update + re-pin caller workflows to pass `app-id`/`private-key` (garden, jackpkgs, zeus×2, hq — and fix hq's stale `toolbox` path).
4. Verify end-to-end.

Refs: #165
ADR: `docs/internal/designs/022-github-app-auth-add-to-project.md`

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -12,6 +12,11 @@ on:
         required: false
         default: 'https://github.com/orgs/ergodicsystems/projects/1'
         type: string
+      owner:
+        description: 'Account that owns the GitHub App installation and the project. Passed to create-github-app-token so the token is minted against this installation regardless of which repo triggers the workflow. Must match where the app is installed.'
+        required: false
+        default: 'ergodicsystems'
+        type: string
       labeled:
         description: 'Optional comma-separated labels; omit or leave empty to add all matching events. When set, items are filtered per label-operator.'
         required: false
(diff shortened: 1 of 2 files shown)


jmmaloney4:
**Pending close — superseded by the webhook-service direction.**

We're moving project auto-add off GitHub Actions entirely: instead of an App token + per-repo workflows/secrets, a small service in the garden k8s cluster will receive GitHub App webhooks and add items to the board centrally (one secret, no per-repo workflow files, new repos covered automatically). Decision: one central board (`ergodicsystems/projects/1`), authed by a single classic PAT held in-cluster (real cross-org items — installation tokens can't read cross-org content, so the App path would degrade cross-org adds to draft links).

Leaving this open as the reference for the interim, but it will be **closed once the webhook service lands**. ADR + plan incoming. Refs #165.

PR #258, opened 2026-06-10
feat(litellm): manage Team/ApiKey as dynamic resources with an update path

jmmaloney4:
## Summary

`LiteLLMTeam` and `LiteLLMApiKey` were `command:local:Command` shell-outs with only `create`/`delete` handlers and **no `triggers`**. Changing a team's `models` (or any other field) never re-ran anything, and even if it had, `create-team` is find-or-create and never calls `/team/update` — so a team's allowed-model list was **frozen at creation time**. Concretely: adding a new model tier (e.g. `local`) to a team's list silently failed to take effect, producing `team not allowed to access model … Tried to access local` at request time.

This rewrites both as Pulumi **`dynamic.ResourceProvider`s** with a full `check`/`diff`/`create`/`update`/`delete` lifecycle — the same in-repo pattern as `r2/r2object.ts` and `d1/d1-query.ts`.

### What changed
- **`diff()` reconciles drift.** Changes to `models`/`maxBudget`/`budgetDuration`/`teamAlias`/`tags`/`metadata` now call `/team/update` (or `/key/update`) in place on the next `pulumi up`. Only an explicit `team_id` change (team), or a key-value rotation / team move (key), forces a replacement.
- **`create()` adopts existing objects.** Find-or-create is preserved; when the team/key already exists it is adopted and reconciled via update — which is what makes a safe state-delete cutover from the old Command resources possible.
- **Transport: in-process k8s port-forward.** The provider opens a short-lived port-forward to a ready proxy pod via `@kubernetes/client-node`, then `fetch()`s the admin API over `localhost`. Same reachability model as the old `kubectl exec` script (works wherever kube creds work; **no tailnet dependency**, CI-safe). All k8s/`node:net` imports are dynamic `import()`s inside the provider callbacks so the closure serializes — the constraint the R2 dynamic provider documents.
- **Drop-in API.** Public classes, args, and outputs (`key`/`tokenId`/`teamId`) are unchanged. The `${name}-secret` `RandomPassword` is preserved, so **existing key values do not rotate**.
- Retires `scripts/litellm-admin.sh`; adds `@kubernetes/client-node` as a dependency.

## Test plan
- `pnpm build` — passes (tsc emit + script copy).
- `tsc -p tsconfig.build.json --noEmit` — clean (validates the `@kubernetes/client-node` 1.4.0 API usage: `readNamespacedDeployment`/`listNamespacedPod` object params, `PortForward`, `makeApiClient`).
[...]

PR #259, opened 2026-06-10
feat(onepassword): OnePasswordItem dynamic write resource (ADR 031)

jmmaloney4:
## Summary

Adds a generic Pulumi **dynamic resource** that **writes** a 1Password item through the in-cluster 1Password **Connect** server, reached over an **in-process Kubernetes port-forward**. This lets out-of-cluster Pulumi runs (e.g. GitHub Actions over the tailnet apiserver) create/update 1Password items **without exposing Connect** beyond `ClusterIP`.

Design: `docs/internal/designs/031-onepassword-item-write-resource.md` (included in this PR). First consumer: garden ADR 091 (publish a LiteLLM virtual key into 1Password).

## What's here

- **`packages/sector7/onepassword/item.ts`** — `OnePasswordItem` (`sector7:onepassword:Item`), a `pulumi.dynamic.ResourceProvider` with full `check`/`diff`/`create`/`update`/`delete`:
  - **Transport:** short-lived in-process port-forward (`@kubernetes/client-node` `PortForward` + `net.createServer`) to a ready Connect pod (pod selector resolved from the Deployment at runtime), then `fetch` over `localhost`. Same model as `litellm/admin.ts` (sector7#258); accepts an explicit `kubeconfig` (falls back to `loadFromDefault()`) so callers whose creds come from a Pulumi stack output can pass it in.
  - **Lifecycle:** `create()` is find-or-create (adopt an existing item by title and reconcile in place, else POST). `diff()` sends a field-value change to an in-place `PUT` and forces replacement **only** on a vault/title (identity) change; never delete-before-replace. `delete()` removes the item.
  - **Secrets:** `connectToken` and `kubeconfig` are marked `additionalSecretOutputs` (encrypted in state); field **values are not echoed into state** — drift is tracked via a `contentHash`.
  - All Node/k8s imports are lazy `await import()` inside the callbacks (closure-serialization rule, per `r2object.ts` / `admin.ts`).
- **`packages/sector7/onepassword/index.ts`** — barrel (explicit named re-exports).
- **`package.json`** — new `./onepassword` subpath export; **`barrel-guard.ts`** asserts it stays off the root barrel (AGENTS.md).
- **`tests/onepassword-item.test.ts`** — mock-based lifecycle tests mirroring `litellm-admin.test.ts` (Pulumi + k8s client + `node:net` + `fetch` stubbed): check (valid / missing), create (new POST vs adopt PUT), diff (no-change / in-place value change / vault & title replacement), update, delete.

## Scope / follow-ups

- The port-forward transport is intentionally kept **in-module** (mirroring `admin.ts`) so the serialized provider closure stays self-contained. Factoring a shared port-forward util shared with `litellm/admin.ts` is a follow-up once both land (noted in ADR 031; `#258` is the other half).
- v1 stores arbitrary fields; the first consumer needs a single `password` field readable as `op://<vault>/<title>/password`.

## Test plan

- `tsc -p tsconfig.json --noEmit` — clean.
- `pnpm run build` — emits `dist/onepassword/`.
- `pnpm test` — **218 passed** (11 new).
- `biome check` — new files formatted.

CI gates here are `tsc` + `vitest` (`flake.nix` `jackpkgs.checks`); both pass locally.

## Risks

- Low-to-moderate: new dynamic provider; no existing code paths changed. The transport depends on apiserver reachability + a ready Connect pod at deploy time (clear errors otherwise). Real `pulumi up` against Connect is exercised by the garden ADR 091 consumer, not this PR.

### DIFF
--- a/docs/internal/designs/031-onepassword-item-write-resource.md
+++ b/docs/internal/designs/031-onepassword-item-write-resource.md
@@ -64,6 +64,7 @@ port-forward**.
 ## Type token & package surface
 
 - Type token: `sector7:onepassword:Item`.
+
 - Subpath export (not the root barrel — it pulls in `@kubernetes/client-node`):
 
   ```ts
--- a/docs/internal/designs/031-onepassword-item-write-resource.md
+++ b/docs/internal/designs/031-onepassword-item-write-resource.md
@@ -1,0 +1,258 @@
+---
+id: ADR-031
+title: OnePassword Item Write Resource
+status: Proposed
+date: 2026-06-10
+deciders: [jmmaloney4]
+consulted: []
+tags: [design, adr, pulumi, kubernetes, onepassword, secrets]
+supersedes: []
+superseded_by: []
+links:
+  - garden ADR 091 (docs/internal/decisions/091-pulumi-writes-litellm-keys-to-1password.md)
+  - sector7#258 (LiteLLM Team/ApiKey as dynamic resources)
+---
+
+# Context
+
+We need a way to **write** (create and update) a 1Password item from Pulumi —
+i.e. to publish a value Pulumi already holds into 1Password so that downstream
+read paths (the 1Password operator's `OnePasswordItem` CR → k8s Secret, or a
+`op read` on a workstation) can distribute it. The triggering consumer is
+**garden ADR 091**: the garden LiteLLM stack mints virtual keys (e.g.
+`personalHermesAgentApiKey`) and wants to store one in 1Password instead of
+leaving it in Pulumi state and hand-copied plaintext.
+
+The forces that shape this:
+
+- **1Password Connect is `ClusterIP`-only** in the target clusters
+  (garden's `deploy/platform/1password/`). We MUST NOT add tailnet/public ingress
+  to Connect (garden ADR 088 keeps these surfaces private). The Connect **server**
+  REST API is full CRUD; the **operator** only syncs items *out* to Secrets, so it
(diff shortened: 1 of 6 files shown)


jmmaloney4:
Re: the ticket-compliance check against #258 — that's a misattribution. This PR is **ADR 031** (a new, generic `OnePasswordItem` dynamic write resource), not an implementation of #258. #258 is referenced only because it (a) established the in-process port-forward transport pattern this resource reuses and (b) is the other half of the eventual shared-transport dedup. It is intentionally **not** trying to satisfy #258's LiteLLM-team/key requirements.

Addressed the three actionable findings in `1ac7f1a`:
- **Secret exposure** — `connectToken`, `kubeconfig`, and every field `value` are now wrapped in `pulumi.secret()` in the constructor, so sensitive inputs are encrypted in state even if a caller passes plain strings (previously only `additionalSecretOutputs` covered the output side).
- **Ambiguous adoption** — adoption now refuses when more than one item shares the title (titles aren't unique in a vault): exactly one match adopts, zero creates, more is a hard error. Same authorization reasoning #258 applies to team aliases.
- **Stale transport state** — `diff()` now treats a transport-input change (token/kubeconfig/namespace/deployment/port) as an in-place change so state is refreshed, rather than silently keeping stale values that would break a later `update()`/`delete()`.

Added tests for both new behaviors; `tsc` + `vitest` (220 passed) + `biome` all clean.

PR #270, opened 2026-06-12
feat(nix): incrementally push built paths to Attic via watch-store

jmmaloney4:
## Summary

Makes the Nix CI cache **seed reliably** and stops the `detect` job from rebuilding nix-eval-jobs every run. Adds an optional `attic watch-store` to `nix-setup` that, when Attic push creds are present, runs in the background for the whole job and uploads every store path **as it's built**.

## Why

Today the only push is a single `push-attic-cache` step at the *end* of the build job. With `concurrency: cancel-in-progress`, a run cancelled mid-build (very common under rapid merges, and made worse by an empty/cold cache) pushes **nothing** — so the cache never seeds. Separately, the `detect` job runs `nix run github:nix-community/nix-eval-jobs/v2.34.1`, whose closure is in no public cache, so it **builds ~11 derivations (nix 2.34.1, curl, nix-eval-jobs) from source every run**.

`attic watch-store` fixes both: it watches `/nix/store` and pushes paths as they appear, so
1. a cancelled build still uploads everything it finished (resilient seeding), and
2. the first `detect` run pushes the nix-eval-jobs closure to Attic → subsequent `detect` runs **substitute** it instead of rebuilding.

## Why watch-store, not a Nix `post-build-hook`

The runner installs Nix via DeterminateSystems (multi-user daemon). A `post-build-hook` runs as **root in the daemon's restricted context** and **fails the build if it errors** — risky for shared CI across all repos. `watch-store` runs as the runner user, needs no root/daemon config, and is strictly **best-effort**: it never fails the job and is a no-op when creds are absent (so non-Attic repos are unaffected).

## Changes

- `nix-setup`: new optional inputs `attic-endpoint` / `attic-cache-name` / `attic-token`; new `start-attic-watch-store.sh` step that logs in and starts `attic watch-store` in the background (only when all three are set).
- `nix.yml`: passes the creds into both the `detect` and `build` `nix-setup` calls; `detect` gains a job-level `BINARY_CACHE_TOKEN`. The end-of-job `push-attic-cache` step stays as a backstop for final outputs.

## Test plan / validation

- `bash -n` + `shellcheck`: clean.
- `actionlint` on `nix.yml`: clean.
- **Needs a real CI run to validate end-to-end** (watch-store start, login, background push, and that `detect`'s nix-eval-jobs closure lands in Attic). Safe by construction: gated on creds, best-effort, never fails the job.

## Rollout note

Consuming repos pin `nix.yml` at a SHA, so they pick up the `nix.yml` change on their next sector7 bump. The `nix-setup` action change is referenced `@main`, but only takes effect once a pinned `nix.yml` passes the new inputs.

### DIFF
--- a/.github/actions/nix-setup/action.yml
+++ b/.github/actions/nix-setup/action.yml
@@ -17,6 +17,18 @@ inputs:
     description: Optional newline-delimited extra trusted public keys to append to the runner Nix config
     required: false
     default: ''
+  attic-endpoint:
+    description: Optional Attic server endpoint (root URL). When set together with attic-cache-name and attic-token, starts `attic watch-store` to push built store paths to the cache incrementally for the rest of the job.
+    required: false
+    default: ''
+  attic-cache-name:
+    description: Optional Attic cache name to push into via watch-store.
+    required: false
+    default: ''
+  attic-token:
+    description: Optional Attic token with push permission, used by watch-store.
+    required: false
+    default: ''
 runs:
   using: composite
   steps:
@@ -37,3 +49,14 @@ runs:
         INPUT_EXTRA_SUBSTITUTERS: ${{ inputs.extra-substituters }}
         INPUT_EXTRA_TRUSTED_PUBLIC_KEYS: ${{ inputs.extra-trusted-public-keys }}
       run: bash "${{ github.action_path }}/configure-binary-cache.sh"
+    # Best-effort incremental cache push: when Attic push creds are provided,
+    # start `attic watch-store` in the background so every store path built in
+    # this job is uploaded as it appears (survives concurrency cancellation and
+    # seeds the detect job's nix-eval-jobs closure). Never fails the job.
+    - if: ${{ inputs.attic-endpoint != '' && inputs.attic-cache-name != '' && inputs.attic-token != '' }}
+      shell: bash
+      env:
+        INPUT_ATTIC_ENDPOINT: ${{ inputs.attic-endpoint }}
+        INPUT_ATTIC_CACHE: ${{ inputs.attic-cache-name }}
+        INPUT_ATTIC_TOKEN: ${{ inputs.attic-token }}
+      run: bash "${{ github.action_path }}/start-attic-watch-store.sh"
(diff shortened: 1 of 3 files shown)


jmmaloney4:
Triage of the security finding (HEAD 521a7ab):

✅ **Token persisted to disk** — addressed: watch-store now logs in under a per-job `XDG_CONFIG_HOME` (mktemp), so the token isn't written into the runner's shared `$HOME` config.

📌 **Token as `attic login` CLI arg (visible in `/proc/<pid>/cmdline`)** — acknowledged, non-blocking: this is inherent to `attic login`'s interface and is **identical to the existing `push-attic-cache` action** already in this repo, so it introduces no new exposure pattern. Our runners are single-tenant ephemeral ARC pods (one job per pod, destroyed at job end), so there is no co-tenant to read `/proc` and no persistence across jobs. If we ever want to eliminate it, that's a follow-up against attic-client (env/stdin token input) affecting both this and `push-attic-cache` uniformly.

Both importance-8 suggestions ("isolate per-job state", "fully detach watcher" via `setsid`/`</dev/null`) are applied in 521a7ab.

jmmaloney4:
Final triage (HEAD 521a7ab):

📌 **"Unpinned dependency" (`nix profile install nixpkgs#attic-client`)** — non-blocking, by convention: the entire workflow resolves tooling via unpinned `nixpkgs#…` (`nixpkgs#jq`, `nixpkgs#nix-fast-build`, and the existing `push-attic-cache` action's `nixpkgs#attic-client`). Pinning only this one would be inconsistent; pinning the toolchain is a separate repo-wide decision.

📌 **"Token exposure"** — addressed (per-job `XDG_CONFIG_HOME`, no shared-home persistence) + documented above (CLI-arg form is inherent to `attic login`, matches `push-attic-cache`, single-tenant ephemeral pods).

✅ `/improve` fresh run: no code suggestions. Static validation (`bash -n`, `shellcheck`, `actionlint`): clean. The failing `add-to-project` check is unrelated project-board automation.

The watch-store path is gated on creds, best-effort, and never fails the job, so merging is low-risk; first end-to-end validation happens on the next downstream CI run after the consumer pin bump.

PR #271, opened 2026-06-12
feat(gateway): support HTTPRoute rule timeouts in createServiceHttpRoute

jmmaloney4:
## Summary

`createServiceHttpRoute()` emits HTTPRoutes with no `timeouts` field, so Envoy Gateway applies Envoy's built-in **15s route timeout**. Any request slower than 15s is killed with a 504 (`upstream request timeout`) — and since the route timeout bounds the *entire* response, it fires even while an SSE stream is actively flowing. This is the root cause of jmmaloney4/garden#803 and the dominant trigger behind jmmaloney4/garden#796: live Envoy access logs on garden's `private-gateway` show a hard wall of `UT`-flagged litellm requests at 15.0–15.8s, and no request through the gateway has ever completed past ~15.8s.

This PR adds an optional `timeouts` arg (Gateway API GEP-1742: `request` / `backendRequest`) to `ServiceHttpRouteArgs`, threaded into `spec.rules[0].timeouts`. Values are validated against the Gateway API duration format (GEP-2257) up front, following the existing `pathPrefix` fail-fast pattern, so a bad duration fails at preview rather than producing an invalid manifest at apply.

No behavior change for existing callers — the field is omitted from the manifest when not passed.

## Test plan

- 4 new vitest cases: type surface accepts `timeouts`, accepts `"0s"` (disable), rejects non-duration strings for both fields (`pnpm test`: 262 passed).
- `pnpm run build` (tsc) passes.
- `nix flake check -L` passes.
- End-to-end validation happens in the follow-up garden PR (litellm callsite + live Envoy log verification).

## Risks

- Low: additive, optional field. The duration regex matches GEP-2257; exotic-but-valid multi-group durations like `"1h30m"` are accepted.
- `backendRequest <= request` is documented but not cross-validated; the gateway controller rejects violations at admission.

### DIFF
--- a/packages/sector7/gateway/index.ts
+++ b/packages/sector7/gateway/index.ts
@@ -29,6 +29,21 @@ export interface HttpRouteTimeouts {
 // unit (h, m, s, ms), e.g. "600s", "1h30m", "0s".
 const GATEWAY_API_DURATION = /^([0-9]{1,5}(h|m|s|ms)){1,4}$/;
 
+const DURATION_UNIT_MS: Record<string, number> = {
+	h: 3_600_000,
+	m: 60_000,
+	s: 1_000,
+	ms: 1,
+};
+
+function gatewayApiDurationToMs(duration: string): number {
+	let total = 0;
+	for (const match of duration.matchAll(/([0-9]{1,5})(ms|h|m|s)/g)) {
+		total += Number(match[1]) * (DURATION_UNIT_MS[match[2] ?? ""] ?? 0);
+	}
+	return total;
+}
+
 export interface ServiceHttpRouteArgs {
 	/**
 	 * Logical/resource name for the HTTPRoute (e.g. "periscope").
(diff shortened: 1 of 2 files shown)


jmmaloney4:
Review triage for `5091f3a` (current head):

**"Invalid validation" (duration regex accepts `1m1m` / `1s1m`)** — Status: false positive, no code change. The regex used here is byte-for-byte the GEP-2257 spec regex, which is also the `Pattern` on the Gateway API CRD `Duration` type (`^([0-9]{1,5}(h|m|s|ms)){1,4}$`). GEP-2257 explicitly defines semantics via Go `time.ParseDuration`, which accepts repeated and unordered unit groups (`1m1m` = 2m). Any value this validator accepts is accepted verbatim by CRD admission, so there is no preview/apply divergence. Enforcing canonical ordering would *reject* spec-valid values. The local ms-converter also sums repeated groups correctly, so the ordering check is sound for these inputs.

**"Make the test meaningful" (`.not.toThrow(/must not exceed/)`)** — Status: acknowledged, no code change. Vitest's negated `toThrow(pattern)` already has exactly the suggested semantics: it fails only if the function throws *and* the message matches the pattern; an unrelated error (e.g. the expected Pulumi-runtime constructor failure noted in the test comment) does not satisfy it. The suggested capture-and-assert rewrite is behaviorally identical.

jmmaloney4:
Status: Already fixed in 5091f3a. The empty-timeouts case (including all-undefined fields like `{ request: undefined }`) is now filtered before emission — the `timeouts` field is omitted from the manifest unless at least one field is set.

PR #284, opened 2026-06-15
ci(pnpm): add attic cache support

jmmaloney4:
## Summary
- add Attic binary cache inputs to the reusable `pnpm.yml` workflow
- thread those inputs through every `nix-setup` call in `analyze`, `verify-lockfile`, and `publish-packages`
- match the cache-selection behavior already used in `pulumi.yml`: keep FlakeHub fallback unless both custom cache URL and public key are provided

## Why
- `pnpm.yml` was the odd workflow out: `nix.yml` and `pulumi.yml` already support custom Attic substituters and push credentials
- this lets callers seed and reuse the same binary cache path for pnpm CI without forking the workflow

## Validation
- `git diff --check`
- `nix flake check -L` *(fails on pre-existing unrelated treefmt churn in `renovate/minor-patch-automerge.json`; the workflow change itself evaluates cleanly and the pnpm deps derivation starts successfully)*

## Risks
- low risk: change is limited to reusable workflow inputs and `nix-setup` wiring
- no behavior change for existing callers unless they pass the new cache inputs/secrets

### DIFF
--- a/.github/workflows/pnpm.yml
+++ b/.github/workflows/pnpm.yml
@@ -25,6 +25,30 @@ on:
         required: false
         type: string
         default: 'release'
+      binary-cache-url:
+        description: 'Optional full binary cache URL to add as a substituter (e.g. https://cache.example.com/cache)'
+        required: false
+        default: ''
+        type: string
+      binary-cache-public-key:
+        description: 'Optional trusted public key for the binary cache'
+        required: false
+        default: ''
+        type: string
+      binary-cache-endpoint:
+        description: 'Optional Attic server endpoint (root URL) used for cache pushes'
+        required: false
+        default: ''
+        type: string
+      binary-cache-name:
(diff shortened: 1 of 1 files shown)


jmmaloney4:
Addressed PR-Agent's actionable cache-wiring findings on `21ddbc45`:
- gate `extra-substituters` / `extra-trusted-public-keys` on complete URL+public-key config
- gate Attic push params on complete endpoint+cache-name+token config

Leaving the low-importance least-privilege suggestion as non-blocking:
- `nix.yml` and `pulumi.yml` already pass Attic push credentials into non-publish jobs to seed/store closures via `nix-setup`'s `attic watch-store` path
- keeping `pnpm.yml` aligned with that pattern is intentional for cache warmup behavior

Local validation remains the same: `git diff --check` passed; `nix flake check -L` is still blocked only by pre-existing unrelated treefmt churn in `renovate/minor-patch-automerge.json`.

PR #296, opened 2026-06-22
feat(nix-setup): use host Nix daemon when present, skip in-pod install

jmmaloney4:
**DRAFT** — PR2 of the warm CI store work (ADR `105-ci-runner-host-nix-daemon-warm-store`, ergodicsystems/hq#173). Spec: jmmaloney4/garden#1007. Pairs with the garden ARC mount PR (PR3). **Do not merge** until validated on a canary runner.

## Why

The slow `detect`/nix-eval-jobs job is a cold-store evaluation — every ephemeral runner installs Nix from scratch. ADR 105's fix is to give runners a **warm shared `/nix`** by bind-mounting the node's store + daemon socket into the pod (garden PR3) and having them use the **host** Nix daemon instead of installing their own. Two independent daemons writing one `/nix` corrupt the store, so the in-pod installer must be skipped when a host daemon is present.

## What

`.github/actions/nix-setup/action.yml`: a new **auto-detecting** gate.

- New `Detect host Nix daemon` step probes `/nix/var/nix/daemon-socket/socket`.
- **When present:** set `NIX_REMOTE=daemon`, put the host store's Nix CLI on `PATH`, and **skip** the in-pod `nix-installer-action`, `flakehub-cache-action`, and client-side `configure-binary-cache.sh`.
- **When absent (cloud runners):** behavior is **unchanged** — installer runs as before.

Auto-detection means no caller change is needed and it safely no-ops until garden PR3 actually mounts the socket.

## Open questions for canary validation

1. **Nix CLI on PATH.** The node mounts `/nix` but not `/run/current-system`. This uses `/nix/var/nix/profiles/default/bin` — needs confirming that path exposes a usable `nix` on voldemort (may need PR3 to also surface the system profile / a wrapper).
2. **Daemon-client trust.** The runner connects to the host daemon as the pod uid; confirm it can build/realise. `trusted-users`/uid mapping is deferred to PR3 — verify whether it's actually needed for the CI workload (plain builds don't require trust; client-side substituter overrides do, which we now skip).
3. **watch-store** step left as-is (best-effort, creds-gated) — confirm it behaves sanely against the host store.

## Validation

`action.yml` parses as valid YAML. Real validation is a canary runner with the PR3 mount — tracked before un-drafting.

### DIFF
--- a/.github/actions/nix-setup/action.yml
+++ b/.github/actions/nix-setup/action.yml
@@ -53,9 +53,13 @@ runs:
           echo "Host Nix daemon detected; using the host store, skipping in-pod install."
           echo "present=true" >> "$GITHUB_OUTPUT"
           echo "NIX_REMOTE=daemon" >> "$GITHUB_ENV"
-          # The node mounts /nix but not /run/current-system, so expose the
-          # host store's Nix CLI via the daemon's own default profile.
-          echo "/nix/var/nix/profiles/default/bin" >> "$GITHUB_PATH"
+          # The node mounts /nix but not /run/current-system, so expose the host
+          # store's Nix CLI via the NixOS *system* profile, which lives under
+          # /nix (so it comes with the mount) and carries nix. NB: the *default*
+          # profile (/nix/var/nix/profiles/default) does NOT contain nix on
+          # NixOS — verified on voldemort (Determinate Nix 3.21.1), where nix is
+          # at /nix/var/nix/profiles/system/sw/bin/nix.
+          echo "/nix/var/nix/profiles/system/sw/bin" >> "$GITHUB_PATH"
         else
           echo "No host Nix daemon; will install Nix in-pod."
           echo "present=false" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


jmmaloney4:
**Canary validation on voldemort (2026-06-22)** — probe pod (`actions-runner-controller:2.333.1` image, uid 1001, hostPath `/nix`, `NIX_REMOTE=daemon`) on the live node:

- ✅ hostPath `/nix` mount works — `actions-runner-system` has no Pod Security enforcement, no PSA change needed.
- ✅ Daemon socket reachable (`0666`) and `nix store ping` works as uid 1001.
- ✅ **Untrusted client builds**: `nix build nixpkgs#hello` succeeded with `Trusted: 0` → no `trusted-users` / uid-mapping work needed for CI builds.
- ✅ Warm store is real: 54,837 store paths already present on the node.
- ⚠️→✅ **Found + fixed a real bug**: nix is **not** on `/nix/var/nix/profiles/default/bin` (absent on NixOS); it's on the **system profile** `/nix/var/nix/profiles/system/sw/bin` (Determinate Nix 3.21.1). PATH corrected.

Remaining before un-draft: a real `detect`-job end-to-end on a dedicated canary scale set (Stage 3).

PR #311, opened 2026-06-26
ci: pin quarto workflow to self-hosted runner

jmmaloney4:
## What

Pins `.github/workflows/quarto.yml` to the `self-hosted` ARC runner fleet.

## Why

The quarto deploy job referenced the repo variable `vars.ACTIONS_RUNS_ON`, which is **unset** in this repo, leaving the job with an empty `runs-on` (effectively broken). The reusable workflows already default `runs-on` to `self-hosted`; this brings the direct quarto workflow in line.

## Changes

- `runs-on: ${{ vars.ACTIONS_RUNS_ON }}` → `runs-on: self-hosted` (job + nix-setup input)
- `cache: ${{ !contains(vars.ACTIONS_RUNS_ON, 'self-hosted') }}` → `cache: false` (self-hosted runners have a persistent Nix store, so the actions cache is unnecessary)

### DIFF
--- a/.github/workflows/quarto.yml
+++ b/.github/workflows/quarto.yml
@@ -11,7 +11,7 @@ permissions: # required by the action
   pull-requests: write
 jobs:
   deploy:
-    runs-on: ${{ vars.ACTIONS_RUNS_ON }}
+    runs-on: self-hosted
     permissions:
       id-token: write
       actions: read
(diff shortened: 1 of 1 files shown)


jmmaloney4:
Convergence: PR-Agent's importance-5 runner-determinism suggestion (use a label list) is moot in this infra — sector7's ARC scope registers a single Linux/x64 self-hosted runner set, so the bare `self-hosted` label targets it deterministically, matching every other workflow in the ecosystem. The public-repo fork-code-execution trade-off (this activates a previously-broken pull_request quarto render on internal runners) was reviewed and explicitly accepted by Jack. Merging as-is.

PR #316, opened 2026-07-02
fix(nix-output): default drvPath change detection so content changes redeploy

jmmaloney4:
## Summary

Fixes a stale-deploy footgun in `NixOutput`: the underlying `command.local.Command` only re-ran when its `triggers` changed, and the only default trigger was the attr *name*. No consumer passes extra triggers, so after the first `pulumi up` the command never re-ran — content changes silently deployed nothing, with the store path served from Pulumi state.

**Observed in production today** on `theoreticaledge.com` (garden): new posts merged to main, `pulumi up -s dev` reported `126 unchanged` and deployed a store path frozen in state from the stack's first up.

### The fix: derivation-path change detection, on by default

Pulumi cannot know which files feed a nix build — but nix can: that is exactly what the drv hash is. New `changeDetection` arg on `NixOutputArgs`:

- **`"drv"` (default):** at program time, evaluate `nix eval --raw <repoRoot>#<nixAttr>.drvPath` and include the result in the command's `triggers`. The drv path covers every transitive build input (sources as nix copies them, lockfiles, flake inputs, toolchain), so the command re-runs **iff** the build would produce a different result, and previews stay clean otherwise. Cost: one flake eval per preview/up (~6s measured on the garden monorepo flake); a no-op `nix build` does the same eval anyway. Eval failure fails the program with a descriptive error including nix's stderr.
- **`"none"`:** legacy attr-only triggering, for consumers with prohibitively slow evals who manage `triggers` manually.

`nixAttr` remains a trigger in both modes; caller `triggers` append after the computed ones. `subOutput` deliberately does not participate in the eval — all outputs share one drv.

### NixImage

`NixImage` (build mode) composes `NixOutput`, so it inherits the fix; its push command already re-triggers off `storePath`, so drv change → rebuild → re-push propagates end to end. Added a `changeDetection` passthrough on `NixImageArgs` for opt-out.

### ADR

ADR-021's "Trigger semantics" section amended: documents the incident, the new default, and the rejected alternatives (source narHash / git SHA — wrong in both directions; always-re-run nonce — correct results but permanent preview diffs and state churn).

## Behavior change / rollout note

This is intentionally a behavior change (correct-by-default). On the first `pulumi up` after upgrading, every `NixOutput`/`NixImage(build)` re-runs once (the trigger set changed) — that one-time re-run is itself the fix deploying. Consumers also start paying one flake eval per preview/up; `changeDetection: "none"` opts out.

## Test plan

- [x] `pnpm -C packages/sector7 test` — 291/291 pass. (`tests/renovate-pulumi.test.ts` fails to collect identically on untouched `main` — pre-existing, unrelated.)
- [x] New tests: default drv trigger present with exact `nix eval` invocation; `changeDetection: "none"` omits it and never execs; trigger ordering `[nixAttr, drv, ...custom]`; `resolveDrvPathTrigger` trims output and surfaces nix stderr on failure; `nix-image` build-mode trigger composition `[nixAttr, drv, imageTag]`.
- [x] `pnpm -C packages/sector7 build` (tsc) clean.
- [x] Shorthand attr resolution + eval cost verified against the real garden flake: `nix eval --raw '<garden>#theoretical-edge-html-dev.drvPath'` → drv path in ~6s.

## Risks

- Consumers with very expensive flake evals see slower previews; mitigated by opt-out.
- The drv eval runs at program construction (synchronous `execFileSync` inside an `apply`); on stacks where `repoRoot`/`nixAttr` are unknown during preview the trigger stays unknown, which Pulumi handles as a potential-change — same fallback semantics as `previewStrategy: "resource"`.

## Links

- ADR-021 (amended in this PR)
- Incident: jmmaloney4/garden theoreticaledge.com dev stack stale deploy, 2026-07-02

### DIFF
--- a/docs/internal/designs/021-nix-output-component.md
+++ b/docs/internal/designs/021-nix-output-component.md
@@ -161,11 +161,48 @@ The `NixOutput` component runs `nix-output-resolve.sh`. The refactored `NixImage
 
 ## Trigger semantics
 
-Default trigger for `NixOutput` is `nixAttr` (the attribute path as a string). This means changing the attribute — e.g. from `packages.x86_64-linux.lens-api-image` to `packages.x86_64-linux.lens-api-v2-image` — will trigger a rebuild.
+> **Amended 2026-07-02.** The original design triggered only on `nixAttr`
+> (plus caller-supplied `triggers`). In practice no consumer passed extra
+> triggers, so `command.local.Command` never re-ran after the first
+> deploy: content changes silently served the stale store path from
+> Pulumi state (observed on the theoreticaledge.com dev stack, where a
+> `pulumi up` after new posts merged reported "unchanged" and deployed
+> nothing). Correct-by-default replaces opt-in triggering.
 
-Consumers who need finer-grained triggering (e.g. flake lock hash, git commit SHA) can pass `triggers` directly.
+`NixOutput` computes a **derivation-path trigger** by default
+(`changeDetection: "drv"`): at program time it runs
+`nix eval --raw <repoRoot>#<nixAttr>.drvPath` and includes the result in
+the child command's `triggers`. The drv path is nix's own hash over every
+transitive build input (sources as nix would copy them, lockfiles, flake
+inputs, toolchain), so:
 
-For `NixImage`, the default trigger remains `imageTag` since that's the deployment-level signal that matters. Additional triggers can be layered in.
+- the command re-runs **iff** the build would produce a different result;
+- `pulumi preview` stays clean when nothing relevant changed;
+- Pulumi never needs to know which files feed the build — nix answers that.
+
+Cost: one flake evaluation per preview/up. Nix's eval is the same work a
+no-op `nix build` would do, so this does not meaningfully slow deploys.
+An eval failure fails the program with a descriptive error (the build
+could not have succeeded either).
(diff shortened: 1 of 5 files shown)


jmmaloney4:
**Triage: "Potential Command Injection" in `resolveDrvPathTrigger` — non-blocking, no change.**

1. No shell is involved (`execFileSync` argv), so there is no injection in the security sense — the review acknowledges this.
2. The constructed flake ref `${repoRoot}#${nixAttr}` is byte-identical to what `nix-output-resolve.sh` has always passed to `nix build`/`nix eval` (`"${REPO_ROOT}#${FULL_ATTR}"`). Any input containing `#` or other ref-special characters misbehaves identically in the pre-existing resolve/build step; the trigger eval adds **no new interpretation surface**.
3. Per ADR-021's security section, `repoRoot`/`nixAttr` are program-controlled Pulumi inputs, not user-facing data.
4. Failure mode is closed: a malformed ref makes `nix eval` error, which throws with the nix stderr and fails the program — it cannot silently trigger against the wrong attribute while building the right one.

Adding validation here but not in the script would create false asymmetry; if we ever want ref validation it belongs in `NixOutputArgs` for both paths, which is out of scope for this fix.

PR #328, opened 2026-07-10
feat(pulumi-config): add op:// secret helpers

jmmaloney4:
## Summary

Add reusable `op://` secret-reference helpers to the `@jmmaloney4/sector7/pulumi-config` subpath export. These utilities detect 1Password CLI `op://` references in Pulumi config values, parse them according to the official 1Password syntax, and create `OnePasswordItem` CRDs that sync secrets into Kubernetes Secrets consumable by pods via `secretKeyRef`.

This extracts the polymorphic API key pattern (raw Pulumi secret OR `op://` 1Password reference) currently implemented locally in `garden`'s LiteLLM stack, making it reusable across all stacks that use the 1Password Connect operator.

## Exports

- **`parseOnePasswordItemReference(opRef, accountKey)`** — parses `op://<vault>/<item>/[section/]<field>` into `{ itemPath, fieldName }` for CRD creation. Strips query parameters (`?attribute=otp`, `?ssh-format=openssh`). Accepts both names and UUIDs. Validates 3-4 path segments.
- **`mergeSecretRefEnvs(...maps)`** — merges multiple secret-ref env var maps into one, returns `undefined` when empty.
- **`createOnePasswordSecretRefs(options)`** — high-level helper: iterates config items, detects `op://` refs via a `keySelector`, creates `OnePasswordItem` CRDs, returns `{ envVarName → { secretName, key } }` map. Supports a `blockRef` guard for items that cannot use `op://` (e.g., z.ai accounts whose quota exporter needs the raw key).

## Usage

```typescript
const providerSecretRefs = createOnePasswordSecretRefs({
  config,
  configKey: "backendAccounts",
  items: directBackendAccounts,
  keySelector: (item, key) => {
    const raw = config.get(`backendAccounts.${key}.apiKey`);
    return raw?.startsWith("op://") ? raw : undefined;
  },
  envVarNaming: (key) => key.toUpperCase().replace(/-/g, "_").concat("_API_KEY"),
  namespace: ns.metadata.name,
  provider,
  dependsOn: [ns],
  blockRef: (item, key) =>
    item.provider === "zai"
      ? `Backend '${key}' (zai) cannot use op://`
      : undefined,
});

// Override account API keys to read from env
for (const [envVar, { secretName, key }] of Object.entries(providerSecretRefs)) {
  accounts[reverseKey(envVar)].apiKey = `os.environ/${envVar}`;
}
```

## `op://` syntax compliance

Parser aligns with the official 1Password CLI secret-reference syntax:
- `op://<vault>/<item>/[section/]<field>` — 3 or 4 path segments
- Query parameters (`?attribute=otp`, `?ssh-format=openssh`) are stripped before parsing (they affect `op read` output format but not the CRD sync target)
- Both human-readable names and UUIDs accepted (Connect operator's `itemPath` accepts either)
- Supported characters: `a-z A-Z 0-9 - _ .` and whitespace
- Rejects <3 segments (missing vault/item/field) and >4 segments (unexpected nesting)

Reference: https://developer.1password.com/docs/cli/secret-references

## Test plan

- [x] `vitest run tests/op-secret-helpers.test.ts` — 16/16 passing
- [x] `tsc -p tsconfig.build.json --noEmit` — clean
- [x] Tests cover: 3-segment refs, 4-segment refs, UUID IDs, query param stripping (`?attribute=otp`, `?ssh-format=openssh`), whitespace in names, segment-count validation, merge semantics (varargs, undefined filtering, empty maps)

## Follow-up

Garden adoption (switching LiteLLM stack from local implementations to `@jmmaloney4/sector7/pulumi-config` imports) will follow after this PR merges and a new sector7 version is published.

### DIFF
--- a/packages/sector7/pulumi-config/index.ts
+++ b/packages/sector7/pulumi-config/index.ts
@@ -6,3 +6,10 @@ export {
 	requireMixedConfig,
 	type SecretFieldsOf,
 } from "./mixed-config.js";
+
+export {
+	createOnePasswordSecretRefs,
+	type CreateOnePasswordSecretRefsOptions,
+	mergeSecretRefEnvs,
+	parseOnePasswordItemReference,
+} from "./op-secret-helpers.js";
--- a/packages/sector7/pulumi-config/op-secret-helpers.ts
+++ b/packages/sector7/pulumi-config/op-secret-helpers.ts
@@ -337,5 +337,6 @@ function toDNS1123Name(name: string): string {
 		.toLowerCase()
 		.replace(/[^a-z0-9-]/g, "-")
 		.replace(/-+/g, "-")
-		.replace(/^-|-$/g, "");
+		.replace(/^-|-$/g, "")
+		.slice(0, 253);
 }
--- a/packages/sector7/pulumi-config/op-secret-helpers.ts
+++ b/packages/sector7/pulumi-config/op-secret-helpers.ts
@@ -7,30 +7,42 @@
  */
 
 import type * as pulumi from "@pulumi/pulumi";
-import type * as k8s from "@pulumi/kubernetes";
+import * as k8s from "@pulumi/kubernetes";
 
 /**
- * Parse a 1Password CLI-style `op://` reference into the CRD `itemPath` and
- * field-name components needed to sync a Kubernetes Secret via the
- * 1Password Connect operator.
+ * Parse a 1Password CLI secret-reference URI (`op://`) into the CRD
+ * `itemPath` and field-name components needed to sync a Kubernetes Secret
+ * via the 1Password Connect operator.
  *
- * Accepts:
- *   op://<vault-uuid>/<item-uuid>/<field-name>
- *   op://<vault-uuid>/<item-uuid>/<section>/<field-name>
+ * Implements the official `op://` syntax:
+ *   op://<vault>/<item>/[section/]<field>
  *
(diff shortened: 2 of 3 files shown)


jmmaloney4:
PR-Agent note on the remaining `section lost` finding:

I verified this against upstream 1Password operator behavior. The operator itself does **not** preserve section-qualified duplicate field labels when syncing a `OnePasswordItem` into a Kubernetes Secret. Upstream issue: `1Password/onepassword-operator#83` documents that for items containing multiple sections with the same field label, the **last matching key/value pair wins**.

That means our helper cannot solve this by returning a different `fieldName` contract without diverging from the actual Secret keys produced by the operator. The correct scope for this PR is:

- match the operator's current normalization behavior exactly;
- fail fast on malformed `op://` refs;
- sanitize names the same way the operator does so `secretKeyRef.key` matches reality;
- detect env-var and secret-name collisions in our own helper.

If we need true section-aware duplicate-field support, that requires an upstream operator enhancement (or a different CRD contract like fieldMapping / section-prefixing), not a local parser tweak in sector7.

Pushed follow-up fixes on current HEAD `d9270d8`:
- reject non-`op://` prefixes
- reject empty path segments from consecutive slashes
- match operator secret-key normalization for whitespace / invalid chars
- detect merge-time env-var collisions
- detect normalized K8s secret-name collisions

PR #336, opened 2026-07-25
fix(nix-image): force local builds of digest-bearing JSONs (always-allow-substitutes=false)

jmmaloney4:
## Problem

Recurring `just deploy` failures pushing nix2container images:

```
writing blob: Patch "https://ghcr.io/v2/...": happened during read:
Digest did not match, expected sha256:852d49..., got sha256:907d3d...
```

### Root cause

nix2container records each layer's tar digest in JSON files built with `runCommandLocal` (`allowSubstitutes = false`), so the recorded digests are always computed against the same store that skopeo's `nix:` transport later re-tars at push time. That invariant is what makes archive-less pushes safe.

**Determinate Nix ships `always-allow-substitutes = true` by default, which overrides `allowSubstitutes = false`** and lets a binary cache (attic) serve layer/image JSONs that were built on a *different* machine. If any layer dependency is not bit-reproducible, the substituted JSON's digests can never match the local store, and every push fails deterministically — rebuilding doesn't help because the poisoned JSON is just re-substituted.

Diagnosed end-to-end on `cavinsresearch/zeus` autokiss deploys from itachi: the shared common-layer JSON was substituted from attic (built by CI, `ultimate: false`), while the local store held a locally-built `google-cloud-bigquery` whose `__pycache__/*.pyc` bytes differ wholesale from CI's build. One divergent package out of 193 paths in the layer → permanent digest mismatch.

## Fix

Pass `--option always-allow-substitutes false` to the image `nix build` in both `nix-output-resolve.sh` (build mode, used by `NixOutput`/`NixImage`) and `nix-image-build-push.sh`. This restores nix2container's invariant regardless of the host's nix configuration.

Cost: none in practice — the option only affects derivations that explicitly opted out of substitution, which are small JSON-generation steps that are cheap by design. Normal substitution of actual build artifacts is unaffected.

## Validation

- `bash -n` on both scripts.
- The equivalent manual fix was verified on itachi: deleting the substituted JSONs and rebuilding with `--option always-allow-substitutes false` produced locally-built JSONs (`ultimate: true`) whose digests match the store, after which `skopeo copy nix:<image.json> dir:...` completes cleanly.

### DIFF
--- a/packages/sector7/scripts/nix-image-build-push.sh
+++ b/packages/sector7/scripts/nix-image-build-push.sh
@@ -106,10 +106,18 @@ if [ "${SCRIPT_MODE}" = "resolve" ]; then
       >"${DIGEST_FILE}"
 else
   # Build the image
+  #
+  # always-allow-substitutes=false: nix2container's digest-bearing JSONs are
+  # built with allowSubstitutes=false so their recorded layer digests always
+  # match the local store that skopeo streams from. Determinate Nix defaults
+  # always-allow-substitutes to true, overriding that — a substituted JSON
+  # built on another machine makes the push fail with "Digest did not match"
+  # whenever a layer dependency is not bit-reproducible.
   echo "--- nix build ---"
   echo "NIX_ATTR: ${NIX_ATTR}"
   echo "REPO_ROOT: ${REPO_ROOT}"
-  nix build "${REPO_ROOT}#${NIX_ATTR}" -o "${RESULT_LINK}" -L
+  nix build "${REPO_ROOT}#${NIX_ATTR}" -o "${RESULT_LINK}" -L \
+    --option always-allow-substitutes false
 
   IMAGE_PATH="nix:./${RESULT_LINK}"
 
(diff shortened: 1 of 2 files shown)


jmmaloney4:
Merging with CI wedged, per the wedged-queue policy. Record of validation:

**CI state**: both pending checks (`add-to-project`, `adr-management / push placeholder to main`) have been queued >35 min on the `self-hosted` label with no runner assigned — all three runners in the `jmmaloney4-sector7-6kwsg` scale set are **offline**. Neither queued check is a code gate (project-board + ADR placeholder automation); the `claude review` jobs completed as skipped.

**Local validation** (aarch64-darwin, PR worktree):
- `bash -n` on both modified scripts: OK; treefmt leaves both scripts unchanged (format-clean).
- `nix flake check -L --keep-going` at PR head `ab921ea` vs base `fedaa63`: **identical results** — `renovate-config` ✅ at both; `tsc`/`vitest` ❌ at both from a pre-existing `pnpm-deps` FOD hash mismatch, `treefmt`/`pre-commit` ❌ at both from pre-existing `renovate/minor-patch-automerge.json` formatting debt. The PR introduces no new failures (it only touches two shell scripts).

**Reviews**: PR-Agent — no issues, no suggestions; Gemini — no feedback; Bugbot disabled.

The change itself was validated end-to-end manually on itachi (see PR body).

Follow-ups for main (pre-existing, not this PR): fix `pnpmDepsHash` for the pnpm-deps FOD and re-format `renovate/minor-patch-automerge.json`.

PR #342, opened 2026-08-01
feat(onepassword): port OnePasswordItem to the provider plugin

jmmaloney4:
## Summary

Second resource family off the dynamic providers, after LiteLLM (#340). `sector7:onepassword:Item` replaces the bare `pulumi-nodejs:dynamic:Resource`; the component API is unchanged.

**Five live instances** — `litellm/prod` (`hermes-litellm-key`, `pr-converge-creds`), `matrix/prod`, `ergon/prod`, `gateway/prod`.

## Behaviour preserved

The parts that protect live secrets, each pinned by a test:

- **Adoption by title refuses ambiguity.** Titles are *not* unique within a vault, so adopting an arbitrary match could overwrite — or later delete — an unrelated hand-created secret in a shared vault. Exactly one match adopts, zero creates, **more than one is a hard error**.
- **The update body merges into the live item**, so unmanaged fields, sections, urls and tags survive. A blind rebuild would silently drop them, which matters most when adopting a pre-existing item.
- **`managedLabels`** lets an update remove fields this resource previously managed and that were dropped from input, without touching fields it never managed. On first adoption nothing is removed, only upserted.
- **Delete tolerates 404**, so `pulumi destroy` is idempotent for an item removed out-of-band.
- **Transport changes are in-place, never a replacement** — but must still flow into state, or a token rotation is dropped and breaks a later update/delete.
- **`contentHash` stays secret**: it is a SHA-256 over the field values, so exposing it would give an offline oracle for guessing them.

## The content hash is the risky part

`contentHash` drives `Diff`. A single byte of divergence from the TypeScript makes **every existing item show a spurious change on the first apply** — and for this resource, "change" means rewriting live secrets.

So it is pinned by a golden vector captured by running the TS `computeContentHash` logic under node:

```
canonical: {"category":"API_CREDENTIAL","fields":[{"label":"alpha","value":"p1","type":"STRING","purpose":"USERNAME"},{"label":"mid","value":"\"quoted\"","type":"CONCEALED","purpose":null},{"label":"zebra","value":"a&b<c>d","type":"CONCEALED","purpose":null}]}
sha256:    59a504868fc682a67d70d9ae2371754aa16eb78bd98b99aa09a9c4e6b10cf538
```

Two genuine Go/JS divergences had to be handled, and the fixture exercises both:

| divergence | fix |
|---|---|
| Go's `encoding/json` escapes `&`, `<`, `>`; JS does not | `Encoder.SetEscapeHTML(false)` — fixture value `a&b<c>d` |
| Go sorts map keys; JS preserves insertion order | canonical form built from **structs**, not maps |

Also noted in the code: the TS sorts labels with `localeCompare`, which is not byte order for non-ASCII. Every label in use today is ASCII, where the two agree; the golden test pins the exact digest either way.

## Confirms a schema claim

The generated `sector7:onepassword:Field` type carries `secret=['value']`, so **nested `provider:"secret"` does reach the schema** — the coarse "mark the whole `fields` property secret" fallback flagged as a risk in the design is not needed.

## Test plan

- [x] `go build ./...`, `go vet ./...` — clean
- [x] `go test -race ./...` — clean across all three packages
- [x] Golden content-hash vector matches the TypeScript byte for byte
- [x] Order-insensitivity and category-defaulting covered
- [x] Merge semantics covered: unmanaged fields survive, declared fields upsert, previously-managed-but-dropped fields are removed, and adoption removes nothing
- [x] `pulumi package get-schema` — `sector7:onepassword:Item` with secrets `connectToken`, `contentHash`, `kubeconfig`, and nested `Field.value`

## Not in this PR

The live migration. Retyping these five instances happens per stack, ordered by ascending blast radius — `gateway` → `ergon` → `matrix` → `litellm` — after the LiteLLM cutover proves the alias mechanism. `litellm/prod`'s two items are last because `litellm-personal-keys` is what every dev machine reads via `op://`.

### DIFF
--- a/provider/onepassword/item.go
+++ b/provider/onepassword/item.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -115,9 +116,46 @@ func (Item) Check(ctx context.Context, req infer.CheckRequest) (infer.CheckRespo
 				seen[f.Label] = true
 			}
 		}
+
+		// Collation guard. contentHash sorts fields by label to build the
+		// canonical JSON, and the TypeScript it must match sorted with
+		// String.prototype.localeCompare (onepassword/item.ts:293) — ICU
+		// collation, not byte order. The two DISAGREE on mixed case:
+		// localeCompare orders ["LiteLLM-Key","api-token"] as
+		// ["api-token","LiteLLM-Key"]; Go's `<` gives the reverse. A divergence
+		// changes the hash, which shows up as a spurious diff and rewrites a
+		// live secret.
+		//
+		// Reimplementing ICU collation is the wrong fix: it is a large surface
+		// (a fuzz of a case-folding approximation still diverged on 6.6% of
+		// pairs, all involving `_` and `.`), and getting it subtly wrong would
+		// change the hash for items that work correctly today.
+		//
+		// Instead, restrict multi-field items to the domain where the two
+		// orders are provably identical. Byte order and localeCompare agree on
+		// every pair of lowercase-kebab labels (verified by fuzzing 382,534
+		// pairs under node: zero divergences). A single-field item has nothing
+		// to sort, so its hash cannot depend on collation at all — which is why
+		// the guard is skipped there, and why the live `LiteLLM-Key` item is
+		// unaffected.
+		if len(args.Fields) > 1 {
+			for i, f := range args.Fields {
+				if f.Label != "" && !kebabLabel.MatchString(f.Label) {
+					fail(fmt.Sprintf("fields[%d].label", i), fmt.Sprintf(
+						"multi-field items must use lowercase-kebab labels (got %q): "+
+							"field order feeds contentHash, and byte order only matches "+
+							"the TypeScript localeCompare sort within that domain", f.Label))
+				}
+			}
+		}
 	}
 	return infer.CheckResponse[ItemArgs]{Inputs: args, Failures: failures}, nil
 }
+
+// kebabLabel is the collation-safe label domain: lowercase alphanumerics and
+// interior hyphens. Matches the lowercase-kebab contract garden's gateway
+// already enforces on the only live multi-field item.
+var kebabLabel = regexp.MustCompile(`^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`)
 
 func (Item) Diff(_ context.Context, req infer.DiffRequest[ItemArgs, ItemState]) (p.DiffResponse, error) {
 	olds, news := req.State, req.Inputs
--- a/provider/onepassword/item.go
+++ b/provider/onepassword/item.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


jmmaloney4:
## Addressed: `asHTTPError` now uses `errors.As`

Agreed and fixed in `cc77d4d`.

The assertion works *today* — `httpx.Client.Do` returns `*Error` unwrapped, and the retry loop passes it through with `return err` rather than wrapping. But that is a property nobody is holding onto: one `fmt.Errorf("…: %w")` added anywhere in that client silently breaks the match, and the failure doesn't look like a bug. It looks like `pulumi destroy` failing on an item that was already removed out of band, which keeps the resource in state and reds every subsequent `up`.

Pinned by `TestAsHTTPErrorSeesThroughWrapping`: a 404 must still be recognised through one and two layers of `%w`, and a transport error must not be mistaken for an HTTP status.

The same bare assertion existed in `attic/cache.go` (`isCacheAlreadyExists` and `Cache.Delete`) — fixed there too, on #343.

`go test -race ./...` green.

jmmaloney4:
## Round 2

### 1. ContentHash collation drift — **fixed** (`eab3cf3`), and it was worse than described

The finding framed this as a non-ASCII risk. The actual trigger is **case**, which is far more likely to show up. Confirmed under node:

```
["LiteLLM-Key","api-token"].sort(localeCompare) => ["api-token","LiteLLM-Key"]
byte order (what Go does)                       => ["LiteLLM-Key","api-token"]   ← diverges

["café","cafe"]                                 => agree
```

So the non-ASCII case the review worried about happens to be safe, and the mixed-case case nobody mentioned is the live hazard. For `OnePasswordItem` a hash divergence isn't a cosmetic diff — it **rewrites a live secret**.

**Reimplementing ICU collation is the wrong fix.** I tried the obvious approximation (fold to lowercase, tiebreak lowercase-before-uppercase) and fuzzed it against `localeCompare`: **19,725 divergences in 300,000 pairs (6.6%)**, all involving `_` and `.`. Shipping a subtly-wrong comparator would change the hash for items that are correct today — causing exactly the failure being prevented.

Instead `Check` now restricts **multi-field** items to the domain where the two orders are provably identical — lowercase-kebab labels — verified by fuzzing **382,534 pairs under node with zero divergences**. A single-field item has nothing to sort, so its hash cannot depend on collation, and the guard is skipped there.

Nothing live is affected:

| item | fields | outcome |
|---|---|---|
| litellm `password` | 1 | unconstrained |
| litellm `LiteLLM-Key` | 1 | unconstrained — the guard would otherwise reject it |
| matrix `credential` | 1 | unconstrained |
| ergon trigger-token | 1 | unconstrained |
| gateway `mcp-consumer-tokens` | 3 (`goose`, `hermes`, `claude-code`) | already lowercase-kebab, enforced by `KEBAB_ID` at `gateway/index.ts:810` |

`TestCheckGuardsCollationUnsafeLabels` pins all five boundary cases.

### 2. Retry on Create — **rejected**, with rationale

> The current logic treats all failures as "don't retry," which may be too strict for transient network errors that happen before item creation.

The client cannot distinguish "5xx before creation" from "timeout after creation succeeded." That's the whole reason the flag exists. Retrying trades a **loud, safe** failure for a **silent, unsafe** one:

- **Don't retry (current):** `pulumi up` fails, you re-run it. If the first POST did succeed, `findItemIDByTitle` adopts the existing item on the next run — which is precisely what the find-or-create path is for.
- **Retry:** a duplicate item lands in the vault holding a live secret, and — as the review itself notes — the *next* run then fails with "ambiguous title" and needs manual cleanup.

The proposed refinement also isn't reachable: by the time `Do` sees a transport error there is no information left about whether the server processed the request.

This matches the rule applied to `/key/generate` and `/team/new`: creating calls are never retried, because a retried POST after a timeout-that-succeeded produces a live credential Pulumi never records.

PR #343, opened 2026-08-01
feat(provider): port the remaining sector7 dynamic providers (Attic, D1, R2)

jmmaloney4:
## Summary

Completes the Attic family — `sector7:attic:Cache` (3 live) and `sector7:attic:Token` (10 live). Stacked on #342; merge that first.

These are the **highest-consequence resources in the whole migration**:

- a spurious `Cache` replace regenerates the NAR-signing keypair and breaks every client's `trusted-public-keys`
- a spurious `Token` replace re-mints host tokens and forces an **agenix re-bootstrap across the NixOS fleet**

So the tests target exactly those failure modes.

## Token: parity proven byte-for-byte

`MintToken` produces the **identical JWT string** to the TypeScript, verified by running its exact logic under node with the same inputs and pinning the result:

```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJnaXRodWItYWN0aW9ucy1jaSIs…fQ.y3o9oLiy2W0p3w0vaeB138pQoNIaGoT7pQW_inA7VaE
```

That covers claim order, base64url-without-padding, and that the HMAC uses the **decoded** secret bytes rather than the base64 text.

**The load-bearing property**: Attic deserialises each permission value as an **integer, not a boolean** — `atticadm` emits `{"r":1,"cc":1,…}`. Emitting `true` makes Attic reject the *whole token* with 401. Asserted against raw payload text, with ungranted flags asserted **absent** rather than `0`.

Also: the resource id is a random opaque handle, **never the token**. Pulumi stores ids in plaintext, so using a bearer credential as the id would leak it into state and every `pulumi stack export`.

And `Diff` replaces on claim inputs *and nothing else* — the baked token/exp/nbf live in state, not inputs, which is what stops a no-op `up` from re-minting and invalidating the copy every consumer already holds.

## Cache: the keypair must survive adoption

- **Adoption on `CacheAlreadyExists` goes through PATCH, never POST**, so the existing keypair survives. The patch body deliberately omits both `keypair` and `store_dir` — asserted.
- **Fails closed** when the immutable `store_dir` cannot be confirmed as matching. Adopting a same-named cache pointing at a different store dir would silently record the desired `storeDir` in state while the remote kept its own, misrepresenting drift. Both the mismatch and the absent cases are tested, including that no PATCH is issued after a refusal.
- **A missing `public_key` is a hard error**, not an empty string — persisting `""` would record an unusable signing key and hide a broken Attic API.
- Delete tolerates 404 so destroy and replacement are idempotent.
- Only `cacheName` and `storeDir` replace; priority, visibility, upstreams and retention are all in-place.

Admin tokens are minted per operation with **least-privilege scopes** (create/update/delete get only the flags they need) and a 5-minute lifetime.

## A workaround that disappears

Attic validates the `Host` header against its allowed-hosts config, so the in-cluster Service FQDN must be sent even though the connection is to `127.0.0.1`. Node's `fetch` treats `Host` as a forbidden header and silently drops it — which is why the TypeScript had to drop down to `node:http` and hand-roll a client.

Go's `net/http` honours `req.Host`, so the shared `httpx` client carries it and `atticFetch` is **not ported at all**. Pinned by a test asserting the FQDN reaches the server.

## Improvement over the original

Minting is now **deterministic**. The TypeScript serialised the caches map in input order; Go map iteration is randomised, so patterns are sorted and permission flags emitted in atticadm's fixed order. `TestMintIsDeterministic` mints 20× and would flake instantly without it. Byte-identity with the TS therefore holds for single-cache tokens (what the parity vector uses); for multi-cache tokens the output is equally valid and no longer input-order dependent.

## Test plan

- [x] `go build`, `go vet`, `go test -race ./...` — clean across all five packages
- [x] Byte-for-byte JWT parity with the TypeScript
- [x] Permission values integer; ungranted flags absent; signature verifies against the decoded secret
- [x] Adoption PATCHes without `keypair`/`store_dir`; refuses on store_dir mismatch **and** on an absent store_dir
- [x] Missing `public_key` errors; delete tolerates 404; Host header is the Service FQDN
- [x] Diff: identity-only replacement for Cache, claim-only for Token, no churn on unchanged inputs
- [x] `pulumi package get-schema` — 5 resources, secrets correct on each

## Not in this PR

The live migration of the 13 Attic resources, which happens per stack after the LiteLLM cutover proves the alias mechanism.

### DIFF
--- a/provider/attic/cache.go
+++ b/provider/attic/cache.go
@@ -2,6 +2,7 @@ package attic
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -301,8 +302,9 @@ func (c Cache) Create(ctx context.Context, req infer.CreateRequest[CacheArgs]) (
 // isCacheAlreadyExists recognises Attic's create-collision response: a 400
 // whose body names CacheAlreadyExists.
 func isCacheAlreadyExists(err error) bool {
-	e, ok := err.(*httpx.Error)
-	return ok && e.Status == http.StatusBadRequest && strings.Contains(e.Body, "CacheAlreadyExists")
+	var e *httpx.Error
+	return errors.As(err, &e) && e.Status == http.StatusBadRequest &&
+		strings.Contains(e.Body, "CacheAlreadyExists")
 }
 
 func (c Cache) Update(ctx context.Context, req infer.UpdateRequest[CacheArgs, CacheState]) (infer.UpdateResponse[CacheState], error) {
@@ -349,7 +351,8 @@ func (c Cache) Delete(ctx context.Context, req infer.DeleteRequest[CacheState]) 
 	if err != nil {
 		// Idempotent delete: a cache already removed out of band means the
 		// desired end state is reached, so do not fail destroy or a replacement.
-		if e, ok := err.(*httpx.Error); ok && e.Status == http.StatusNotFound {
+		var e *httpx.Error
+		if errors.As(err, &e) && e.Status == http.StatusNotFound {
 			return infer.DeleteResponse{}, nil
 		}
 		return infer.DeleteResponse{}, err
--- a/provider/attic/cache.go
+++ b/provider/attic/cache.go
@@ -1,0 +1,358 @@
+package attic
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	p "github.com/pulumi/pulumi-go-provider"
+	"github.com/pulumi/pulumi-go-provider/infer"
+
+	"github.com/jmmaloney4/sector7/provider/internal/diffutil"
+	"github.com/jmmaloney4/sector7/provider/internal/httpx"
+	"github.com/jmmaloney4/sector7/provider/internal/kube"
+)
+
+const (
+	defaultDeployment = "attic"
+	defaultPort       = 8080
+	defaultStoreDir   = "/nix/store"
+)
+
+// Least-privilege admin token scopes, minted per operation. The token is
+// self-contained (no server contact) and scoped to the single cache being
+// operated on.
+var (
+	adminCreateFlags = CachePermissionFlags{Pull: true, CreateCache: true, ConfigureCache: true, ConfigureCacheRetention: true}
+	adminUpdateFlags = CachePermissionFlags{Pull: true, ConfigureCache: true, ConfigureCacheRetention: true}
+	adminDeleteFlags = CachePermissionFlags{Pull: true, DestroyCache: true}
+)
+
+// Cache is an Attic binary cache, administered through the cache-config API
+// reached by in-cluster port-forward.
+type Cache struct {
+	Transport kube.Transport
+}
+
+type CacheArgs struct {
+	Namespace string `pulumi:"namespace"`
+	// HS256SecretBase64 is the server's signing secret; admin tokens are minted
+	// from it locally.
+	HS256SecretBase64     string   `pulumi:"hs256SecretBase64" provider:"secret"`
+	DeploymentName        string   `pulumi:"deploymentName,optional"`
+	Port                  int      `pulumi:"port,optional"`
+	CacheName             string   `pulumi:"cacheName"`
+	IsPublic              bool     `pulumi:"isPublic,optional"`
+	Priority              int      `pulumi:"priority,optional"`
+	StoreDir              string   `pulumi:"storeDir,optional"`
+	UpstreamCacheKeyNames []string `pulumi:"upstreamCacheKeyNames,optional"`
+	// RetentionPeriodSeconds unset means the server default ("Global").
+	RetentionPeriodSeconds *int `pulumi:"retentionPeriodSeconds,optional"`
+}
+
+type CacheState struct {
+	CacheArgs
+	// PublicKey is the NAR-signing key, read back from the cache config. Every
+	// client's trusted-public-keys depends on it, which is why adoption must
+	// never regenerate the keypair.
+	PublicKey string `pulumi:"publicKey"`
+}
+
+func (a *CacheArgs) Annotate(ann infer.Annotator) {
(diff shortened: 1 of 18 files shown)


jmmaloney4:
## Scope grew: now Attic + D1 + R2

Pushed two more families onto this branch. The provider now covers **seven resources across five families**:

```
sector7:attic:Cache          secret=[hs256SecretBase64]
sector7:attic:Token          secret=[hs256SecretBase64, token]
sector7:d1:Query             secret=[apiToken]
sector7:litellm:KeyRecord    secret=[keyValue, masterKey, tokenId]
sector7:litellm:TeamRecord   secret=[masterKey]
sector7:onepassword:Item     secret=[connectToken, contentHash, kubeconfig]
sector7:r2:ZoneCachePurge    secret=[apiToken]
```

`go test -race ./...` clean across all six packages.

### D1Query

Both Cloudflare resources talk to a public API, so neither needs the port-forward transport, and **both have zero live instances** — pure code, no migration.

The non-obvious failure mode is preserved: Cloudflare answers **HTTP 200 with `success: false`** for query-level failures, so a 2xx is not sufficient. Getting that wrong would silently record a schema migration that never ran. Every error message is joined, not just the first.

SQL and database identity replace (re-running the statement); a token rotation re-executes in place. Delete is a no-op — schema data outlives the resource.

### ZoneCachePurge

Keeps the validation that prevents a destructive surprise: an **explicitly empty** `files`/`hosts` list is *rejected* rather than silently treated as "purge everything". Omitting the property is how you purge a whole zone. Cloudflare's 30-item cap and the files/hosts mutual exclusion are enforced, `hosts` wins over `files`, and file **ordering is not a change** — otherwise every apply re-purges.

### One latent bug fixed rather than ported

The resource id truncated the trigger with JS `slice(0, 12)`, which counts UTF-16 code units. Naive Go byte slicing would **split a multibyte character and put invalid UTF-8 in a resource id**. Truncation is by rune, with a test using a multibyte trigger asserting no replacement character appears.

Triggers are content hashes in practice, so this was never going to fire — but a resource id is not something that should be corruptible by an unusual input.

### Still not ported

`r2:Object` (needs SigV4 signing) and garden's two Matrix resources.

jmmaloney4:
## All eight sector7 dynamic providers are now ported

Added `r2:Object`, completing the set:

```
sector7:attic:Cache          secret=[hs256SecretBase64]
sector7:attic:Token          secret=[hs256SecretBase64, token]
sector7:d1:Query             secret=[apiToken]
sector7:litellm:KeyRecord    secret=[keyValue, masterKey, tokenId]
sector7:litellm:TeamRecord   secret=[masterKey]
sector7:onepassword:Item     secret=[connectToken, contentHash, kubeconfig]
sector7:r2:Object            secret=[accessKeyId, secretAccessKey]
sector7:r2:ZoneCachePurge    secret=[apiToken]
```

`go build`, `go vet`, `go test -race ./...` clean across all seven packages.

### SigV4 is hand-rolled, and proven

Not pulled from `aws-sdk-go-v2`. The surface needed is deliberately tiny — PUT and DELETE, no query strings, no multipart, no chunked encoding, fixed region `auto` and service `s3` — so ~60 lines avoids a very large dependency for one signer.

Correctness is **proven, not assumed**: a golden vector pins the exact `Authorization` header against the TypeScript, captured by running its logic under node with a fixed date, credentials and body.

SigV4 is unforgiving. One byte off in the canonical request, the header ordering, the path encoding or the key-derivation chain yields a signature R2 rejects with an opaque **403** — so a golden vector is the only way to be confident without a live bucket.

The fixture key contains both a space and a `!`, because they percent-encode differently across implementations: the canonical request must use the **already-encoded** path (JS `URL.pathname`, Go `URL.EscapedPath()`), not a re-encoded one. `content-type` is signed only when present, so signed headers always match sent headers exactly; a DELETE has neither.

### One behaviour deliberately tightened

**`filePath` must now be absolute.** The dynamic provider ran inside the Pulumi Node host, so a relative path resolved against the program directory. A plugin is a separate process with an unspecified working directory, so the same input would resolve somewhere else — or nowhere — **silently**. `uploadAssets` should `path.resolve()` first.

This is the one place the port is stricter than the original, and it is a silent-breakage trap rather than a preference.

### Preserved

A missing file still yields an empty MD5 rather than an error, so `Diff` can run during preview on a file a prior build step has not produced yet — `Check` is where a missing file is reported. Delete tolerates both 204 and 404. ETag quotes are stripped so the stored value compares directly against the local MD5.

### Remaining in the overall effort

Only garden's two Matrix resources (`MatrixBotAccount`, `MatrixRoom`, 4 live), which are garden-local rather than sector7 — a separate PR against a different repo.

jmmaloney4:
## Addressed: the object key is now percent-encoded

Agreed and fixed in `2f32cef`. This one was sharper than it looked, because `objectURL` is **both** the request target and the signing input — so a raw key doesn't just risk a malformed URL, it can sign one object and write to another.

`escapeKey` splits on `/` and escapes each segment. Escaping the whole key would have been wrong in the other direction: it would turn `assets/app.css` into a single segment and move every prefixed object. The bucket name is escaped too.

`TestObjectURLEscapesTheKey` pins six shapes (`a b.txt`, `weird#frag.txt`, `q?uery.txt`, `100%.txt`, `dir/sub dir/f#1.txt`, and the slash-preserving case) and for each asserts three things:

1. the exact expected URL,
2. the URL parses with an **empty query and fragment** — this is the "key leaked into the request line" failure,
3. `url.URL.EscapedPath()` — what the signer canonicalises — equals the path actually sent. A divergence there is a 403 `SignatureDoesNotMatch`, not a visible bug.

`R2Object` has **zero live instances**, so this is a pure correctness fix with no migration surface.

Also carried the `errors.As` fix from #342 into `attic/cache.go`, which had the identical bare `err.(*httpx.Error)` assertion in `isCacheAlreadyExists` and `Cache.Delete`.

`go test -race ./...` green across all 7 packages.

PR #344, opened 2026-08-01
feat(matrix): port MatrixBotAccount and MatrixRoom to the provider plugin

jmmaloney4:
Stacked on #343. Completes the resource port — all **10** dynamic providers now have plugin equivalents:

```
sector7:attic:Cache        sector7:litellm:KeyRecord    sector7:matrix:Room
sector7:attic:Token        sector7:litellm:TeamRecord   sector7:onepassword:Item
sector7:d1:Query           sector7:matrix:BotAccount    sector7:r2:Object
                                                        sector7:r2:ZoneCachePurge
```

## Why these live here

`MatrixBotAccount` and `MatrixRoom` were garden-**local** dynamic providers (`garden/deploy/services/matrix/matrix-{bot-account,room}.ts`), not sector7 ones. They are folded into this provider deliberately: they carry exactly the same serialised-closure fragility that motivated the whole migration, and a second plugin binary would duplicate the packaging, discovery and release machinery for two resources.

Tradeoff accepted: two garden-specific resources land in a shared library.

## Behaviour ported verbatim

Including the parts that look like bugs — this release retypes, it does not fix.

- **`BotAccount.Create` recovers from `M_USER_IN_USE`** by logging in with the same password. That is the half-completed-run path: registration succeeded on the homeserver but Pulumi never persisted state. Failing there would strand the account permanently outside Pulumi.
- **The Read status sets differ between the two resources and are preserved.** `BotAccount` treats 401/404 as gone; `Room` treats 403/404. In particular Room does *not* treat 401 as gone — that means a bad token, not a missing room, and dropping the resource would create a **duplicate** room rather than recover the original.
- **Delete is best-effort on both.** A homeserver that will not deactivate an account, or will not let a bot leave a room, must not block `pulumi destroy` on the rest of the stack. `forget` is skipped when `leave` fails, since forget only works after a successful leave.
- **`Room.Update` invites only newly added members.** Matrix has no "uninvite", so removing someone from the declared list stays a deliberate no-op rather than becoming an error.

## Two things deliberately NOT carried over

**`preset` gets no schema default.** `createRoom` falls back to `private_chat`, but the dynamic provider applied that fallback only when building the request body — it never wrote it into inputs, so live state has no `preset` at all. Declaring `SetDefault` would materialise `preset: "private_chat"` at Check, and Diff would then see a change on an immutable property and **replace all four live rooms**. The fallback stays in `Create`, where it was. Pinned by `TestRoomDiffIgnoresAbsentPreset`.

**`accessToken` is now marked secret.** The dynamic provider did not mark it, so it sits in plaintext in `matrix/prod` state today. Marking it only fixes new writes — the token still wants rotating separately, alongside the LiteLLM master key.

## Non-idempotent calls

Neither `POST /register` nor `POST /createRoom` is retried. A retried POST after a timeout that actually succeeded leaves an account or a room the caller cannot address. Both are pinned by tests asserting exactly one attempt against a 500.

## Known limitation, faithful to the original

When `password` is unset, `Create` generates one but never persists it (it is an input, and outputs spread the inputs). The `M_USER_IN_USE` recovery path would therefore log in with a *different* random password and fail. This is true of the TypeScript today; the Go error message now says so explicitly and points at manual recovery.

## Verification

```
gofmt -l .            # clean
go build ./... && go vet ./...
go test -race ./...   # all packages pass
pulumi package get-schema ./pulumi-resource-sector7 | jq -r '.resources|keys[]'   # 10 resources
```

The Matrix resources need no port-forward or fake transport — `homeserverUrl` is an input, so the tests point it straight at an `httptest.NewServer`. Request bodies are recorded under a mutex; the harness is race-clean.

## Not in this PR

The garden-side TypeScript wrappers and the alias migration for the 4 live Matrix resources. That is a garden PR, and it is sequenced last (Phase N+4) behind litellm/prod.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -107,6 +107,35 @@
           meta.mainProgram = "pulumi-resource-sector7";
         };
 
+        # `go test ./...` across every resource package.
+        #
+        # A separate derivation because the package above sets
+        # subPackages = ["cmd/..."], which also scopes buildGoModule's
+        # checkPhase — so building the plugin proves it COMPILES and nothing
+        # more. These tests are the entire safety argument for retyping
+        # resources that own six live LiteLLM credentials and thirteen Attic
+        # host tokens, so they have to run somewhere CI can see them.
+        #
+        # compute-flake-build-matrix selects checks.* as well as packages.*
+        # (.github/actions/compute-flake-build-matrix/select.nix), so this runs
+        # on every PR with no workflow change.
+        checks.provider-go-test = pkgs.buildGoModule {
+          pname = "provider-go-test";
+          inherit (config.packages.pulumi-resource-sector7) version src modRoot vendorHash;
+          # -race catches the class of bug that has already bitten this
+          # provider twice during the port: a shared map in the port-forward
+          # transport, and an append from the test server's goroutine.
+          buildPhase = ''
+            runHook preBuild
+            go test -race ./...
+            runHook postBuild
+          '';
+          installPhase = "touch $out";
+          # Nothing is installed, so there is no binary to strip or fix up.
+          dontStrip = true;
+          dontFixup = true;
+        };
+
         devShells.default = pkgs.mkShell {
           inputsFrom = [
             config.jackpkgs.outputs.devShell
--- a/flake.nix
+++ b/flake.nix
@@ -127,6 +127,15 @@
           # transport, and an append from the test server's goroutine.
           buildPhase = ''
             runHook preBuild
+            # treefmt has no Go formatter wired up, so gofmt drift reaches main
+            # unchallenged — it already has once.
+            # vendor/ is materialised by buildGoModule and is not ours.
+            unformatted=$(gofmt -l . | grep -v '^vendor/' || true)
+            if [ -n "$unformatted" ]; then
+              echo "gofmt would rewrite:" >&2
+              echo "$unformatted" >&2
+              exit 1
+            fi
             go test -race ./...
             runHook postBuild
           '';
--- a/provider/matrix/botaccount.go
+++ b/provider/matrix/botaccount.go
@@ -1,0 +1,247 @@
+// Package matrix implements Matrix homeserver resources.
+//
+// Ported from garden's deploy/services/matrix/{matrix-bot-account,matrix-room}.ts,
+// which were garden-LOCAL dynamic providers rather than sector7 ones. They are
+// folded into this provider deliberately: they carry exactly the same
+// serialised-closure fragility, and a second provider binary would duplicate the
+// packaging, discovery and release machinery for two resources.
+package matrix
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	p "github.com/pulumi/pulumi-go-provider"
+	"github.com/pulumi/pulumi-go-provider/infer"
+
+	"github.com/jmmaloney4/sector7/provider/internal/httpx"
+)
+
+// BotAccount is a Matrix bot user: registered via the registration API, and
+// deactivated on destroy.
+type BotAccount struct{}
+
+type BotAccountArgs struct {
+	HomeserverURL string `pulumi:"homeserverUrl"`
+	Username      string `pulumi:"username"`
+	DisplayName   string `pulumi:"displayName,optional"`
+	// RegistrationToken is consumed only at registration time, which is why a
+	// change to it forces replacement rather than an update.
+	RegistrationToken string `pulumi:"registrationToken" provider:"secret"`
+	Password          string `pulumi:"password,optional" provider:"secret"`
+}
+
+type BotAccountState struct {
+	BotAccountArgs
+	UserID string `pulumi:"userId"`
+	// AccessToken was NOT marked secret by the garden-local dynamic provider,
+	// so it sits in plaintext in matrix/prod state today. Marking it fixes new
+	// writes; the token itself should be rotated separately.
+	AccessToken string `pulumi:"accessToken" provider:"secret"`
+}
+
+func client(base string) *httpx.Client {
+	return &httpx.Client{
+		BaseURL:    strings.TrimRight(base, "/"),
+		HTTP:       &http.Client{Transport: &http.Transport{DisableKeepAlives: true}, Timeout: 60 * time.Second},
+		MaxRetries: 3,
+	}
+}
+
+func (BotAccount) Check(ctx context.Context, req infer.CheckRequest) (infer.CheckResponse[BotAccountArgs], error) {
+	args, failures, err := infer.DefaultCheck[BotAccountArgs](ctx, req.NewInputs)
+	if err != nil {
+		return infer.CheckResponse[BotAccountArgs]{Inputs: args, Failures: failures}, err
+	}
+	// The original provider validated nothing. These are the fields whose
+	// absence produces a confusing homeserver error rather than a clear one.
+	for _, f := range []struct{ name, val string }{
+		{"homeserverUrl", args.HomeserverURL},
+		{"username", args.Username},
+		{"registrationToken", args.RegistrationToken},
+	} {
+		if f.val == "" {
+			failures = append(failures, p.CheckFailure{Property: f.name, Reason: f.name + " is required"})
+		}
+	}
+	return infer.CheckResponse[BotAccountArgs]{Inputs: args, Failures: failures}, nil
+}
+
+func (BotAccount) Diff(_ context.Context, req infer.DiffRequest[BotAccountArgs, BotAccountState]) (p.DiffResponse, error) {
+	olds, news := req.State, req.Inputs
+	diffs := map[string]p.PropertyDiff{}
+
+	// A different username or homeserver is a fundamentally different account.
+	// The registration token is consumed at registration, so changing it also
+	// requires re-registering.
+	if olds.Username != news.Username {
+		diffs["username"] = p.PropertyDiff{Kind: p.UpdateReplace}
+	}
+	if olds.HomeserverURL != news.HomeserverURL {
+		diffs["homeserverUrl"] = p.PropertyDiff{Kind: p.UpdateReplace}
+	}
+	if olds.RegistrationToken != news.RegistrationToken {
+		diffs["registrationToken"] = p.PropertyDiff{Kind: p.UpdateReplace}
+	}
+	// The display name is mutable in place.
+	if olds.DisplayName != news.DisplayName {
+		diffs["displayName"] = p.PropertyDiff{Kind: p.Update}
+	}
+
+	return p.DiffResponse{
+		HasChanges:          len(diffs) > 0,
+		DetailedDiff:        diffs,
+		DeleteBeforeReplace: false,
+	}, nil
+}
+
+type loginResult struct {
+	UserID      string `json:"user_id"`
+	AccessToken string `json:"access_token"`
+	DeviceID    string `json:"device_id"`
+}
+
+func (b BotAccount) Create(ctx context.Context, req infer.CreateRequest[BotAccountArgs]) (infer.CreateResponse[BotAccountState], error) {
+	out := infer.CreateResponse[BotAccountState]{Output: BotAccountState{BotAccountArgs: req.Inputs}}
+	if req.DryRun {
+		return out, nil
+	}
+	a := req.Inputs
+	c := client(a.HomeserverURL)
+
+	password := a.Password
+	if password == "" {
+		var buf [16]byte
+		if _, err := rand.Read(buf[:]); err != nil {
+			return out, fmt.Errorf("sector7: generating bot password: %w", err)
+		}
+		password = hex.EncodeToString(buf[:])
+	}
+
+	var res loginResult
+	// Never retried: a retried registration after a timeout that actually
+	// succeeded hits M_USER_IN_USE, which the recovery path below handles —
+	// but the register call itself is not idempotent.
+	err := c.Do(ctx, "POST", "/_matrix/client/v3/register", map[string]any{
+		"username": a.Username,
+		"auth": map[string]any{
+			"type":  "m.login.registration_token",
+			"token": a.RegistrationToken,
+		},
+		"password": password,
+	}, &res, false)
+
+	if err != nil {
+		// Recover from a half-completed prior run: registration succeeded on
+		// the homeserver but Pulumi failed to persist state (transient network
+		// error). Log in with the same password to recover the access token
+		// rather than failing forever on a user that already exists.
+		if !isUserInUse(err) {
+			return out, err
+		}
+		if err := c.Do(ctx, "POST", "/_matrix/client/v3/login", map[string]any{
+			"type":       "m.login.password",
+			"identifier": map[string]any{"type": "m.id.user", "user": a.Username},
+			"password":   password,
+		}, &res, false); err != nil {
+			return out, fmt.Errorf(
+				"sector7: Matrix user %s already exists but password login failed — "+
+					"the account exists with a different password and must be recovered manually: %w", a.Username, err)
+		}
+	}
+
+	if a.DisplayName != "" {
+		b.setDisplayName(ctx, c, res.UserID, res.AccessToken, a.DisplayName)
+	}
+
+	out.ID = res.UserID
+	out.Output = BotAccountState{BotAccountArgs: a, UserID: res.UserID, AccessToken: res.AccessToken}
+	return out, nil
+}
+
+func isUserInUse(err error) bool {
+	e, ok := err.(*httpx.Error)
+	return ok && strings.Contains(e.Body, "M_USER_IN_USE")
+}
+
+// setDisplayName is best-effort: a bot with the wrong display name is cosmetic,
+// and failing the whole create over it would strand a registered account
+// outside Pulumi state.
+func (BotAccount) setDisplayName(ctx context.Context, c *httpx.Client, userID, token, name string) {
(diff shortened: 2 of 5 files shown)


jmmaloney4:
## Added: `checks.provider-go-test`

Found while checking why this PR only registered two status checks: **the Go tests were not running in CI at all**, across the whole stack.

`packages.pulumi-resource-sector7` sets `subPackages = ["cmd/pulumi-resource-sector7"]`, which also scopes `buildGoModule`'s `checkPhase`. So building the plugin proved it *compiles* and nothing more — `litellm/`, `attic/`, `onepassword/`, `r2/`, `d1/`, `matrix/` and `internal/kube` never ran a test on a runner. Locally-green is not an argument for a change that retypes resources owning six live LiteLLM credentials and thirteen Attic host tokens.

`checks.provider-go-test` runs the full suite with `-race`. The flag is not decorative — this port has already produced two data races (the port-forward transport's shared config map, and an append from the test server's goroutine).

No workflow change needed: `compute-flake-build-matrix` already selects `checks.*` alongside `packages.*` (`.github/actions/compute-flake-build-matrix/select.nix`).

Verified in **both** directions:

```
$ nix build .#checks.x86_64-linux.provider-go-test
ok  …/provider/attic          1.118s
ok  …/provider/d1             1.085s
ok  …/provider/internal/kube  1.036s
ok  …/provider/litellm        2.788s
ok  …/provider/matrix         4.266s
ok  …/provider/onepassword    1.104s
ok  …/provider/r2             1.091s

# with a deliberately failing test added:
--- FAIL: TestDeliberateFailure (0.00s)
FAIL
```

A check that cannot fail is worse than no check, hence the negative control.

## Note on CI for this PR

`_dogfood-nix.yml` only triggers on PRs targeting `main`, so while this PR is based on `feat/provider-attic-token` it gets no flake checks. GitHub will retarget it to `main` when #343 merges, and the full suite runs then. Until that happens the evidence is local: `gofmt -l` clean, `go build`/`go vet` clean, `go test -race ./...` green, `nix build .#pulumi-resource-sector7` green, and `pulumi package get-schema` emitting all 10 resources.

jmmaloney4:
## Addressed: `password` is now a required input

The finding is **real**, but the suggested fix (`a.Password = password` before building the output state) does not fix it.

The premise of the described scenario is *"the first run registered the account but failed to persist state."* If state was never persisted, the output that fix writes to is discarded along with everything else — run two still generates a fresh random password, still hits `M_USER_IN_USE`, still fails to log in. The one-liner moves the value into a field that, in exactly the failing case, is never saved.

The actual defect is upstream of that: **a generated password is unrecoverable by construction.** It is written nowhere — not to state, not to the outputs, not to 1Password — so it exists only inside one `Create` call. The `M_USER_IN_USE` recovery path is therefore a lie whenever it is taken. And persisting it isn't the answer either: that puts a credential the program never declared into Pulumi state, which is a worse outcome than the one being fixed.

So the fallback is gone and `password` is a required input (`2aa2406`).

Why this is safe:

- All four live call sites in `deploy/services/matrix/index.ts` already pass an explicit `RandomPassword` (e.g. `alerts-bot` at :424). The generated-password path was **dead code in practice**.
- For those resources the recovery path actually works, because the `RandomPassword` value is stable in state across invocations — which is the property the generated password lacked.
- `requiredInputs` is schema metadata; it changes `Check` validation, not property values. Live state already carries `password`, so this is **diff-neutral** for the alias migration.

Net effect: an unrecoverable credential silently minted on the homeserver becomes an up-front `Check` failure. Pinned by `TestBotAccountCheckRequiresPassword`.

## Also in this push

`checks.provider-go-test` now gates on `gofmt -l` as well. treefmt has no Go formatter wired up in this repo, so gofmt drift reaches `main` unchallenged — and it already has once (a stray blank line in `provider/litellm/key_test.go`, since fixed on main). `vendor/` is excluded, being materialised by `buildGoModule` rather than committed.

Verified: 10 resources in the schema, `requiredInputs` for `sector7:matrix:BotAccount` is now `["homeserverUrl","username","registrationToken","password"]`, and `nix build .#checks.x86_64-linux.provider-go-test` is green across all 7 test packages.

## Note on the rebase

This branch was rebased onto the updated #343 to pick up the `errors.As` and R2 key-escaping fixes, so its SHAs changed. Merge order is unchanged: #342 → #343 → #344.

jmmaloney4:
## Round 3 — all three drift findings fixed (`04c4beb`)

Good catches. All three are the same failure shape: an input `Create` sends that no `Update` path can change, and that `Diff` ignored — so the new value lands in Pulumi state while the homeserver keeps the old one, and nothing ever reconciles them. All three were inherited from the dynamic provider.

**`password` → `UpdateReplace`.** This one is sharper than it looks in isolation, because this PR just made `password` load-bearing for the `M_USER_IN_USE` recovery path. Silently accepting a rotation would put a password in state that the account does not have, and recovery would then log in with credentials that do not match. Matrix has a change-password API, but wiring it up is a feature, not a reason to no-op.

**`isDirect` → `UpdateReplace`.** It exists only in the `createRoom` body and no state event can change it, exactly like `preset`. Comparison is nil-safe via `boolPtrEqual` — live rooms have no `isDirect` at all, so absent-vs-absent must compare equal or every one of them gets replaced on the first apply. Same trap as the `preset` default caught earlier in this PR.

**Clearing `displayName` now reaches the homeserver.** `Diff` reported it as an in-place update while `Update` guarded on a non-empty value and skipped the call.

### Plus a latent bug the password finding exposed

The dynamic provider always used `deleteBeforeReplace: false`. But a replacement that reuses the **same homeserver and username** collides with the account still registered there: `/register` answers `M_USER_IN_USE`, and the recovery path then tries the **new** password against an account holding the **old** one — so the replacement fails and the account is stranded.

`DeleteBeforeReplace` is now computed rather than constant:

```go
sameAccount := olds.Username == news.Username && olds.HomeserverURL == news.HomeserverURL
```

true when the replacement targets the same account, false for a username or homeserver change, which by definition targets a different one. This was already reachable through `registrationToken` before `password` joined the diff.

### Tests

`TestBotAccountDiffCatchesPasswordDrift` (including both `DeleteBeforeReplace` branches), `TestBotAccountUpdateClearsDisplayName` (asserts the empty value is actually sent, and that an *unchanged* name still makes no call), `TestRoomDiffTreatsIsDirectAsImmutable` (set, absent-vs-absent, equal-vs-equal, and true→false).

`gofmt` clean, `go vet` clean, `go test -race ./...` green across all 7 packages.

jmmaloney4:
## Round 4 (`d674686`)

### 1. Missing 401 handling in `Room.Read` — **fixed**, though not by reclassifying

The classification is right and stays: a 401 on a room means a bad token, not a missing room, and dropping the resource would create a **duplicate** room rather than recover the original.

But the underlying concern is correct, and it was sharper than the classification argument: `Read` was **discarding the error entirely**. A permanently revoked token left `pulumi refresh` reporting "up to date" forever while every later `Update` or `Delete` failed against a token that no longer works — silent, and exactly the drift described.

Both `Read` paths now log a warning naming the resource and the error:

```
could not verify Matrix room !abc:example.org; keeping existing state: sector7: GET … failed: 401
```

Refresh still does not churn a live room or account over a transient blip — that failure mode is the reason this whole migration exists — but an unverifiable resource no longer masquerades as healthy. Returning the error instead would reintroduce the `pulumi refresh` fragility we are migrating away from.

### 2. Idempotency risk on Create — **already handled**, by the round-2 fix

> if a transient network error occurs *after* the homeserver accepts the registration but *before* the response is received, the retry logic is skipped … stranding an account on the homeserver that Pulumi cannot track

It self-heals on the next run, and deliberately so:

1. **Run 1** — homeserver registers the account, response is lost. `Do` returns a transport error, not `M_USER_IN_USE`, so `Create` fails and the apply fails. Not retrying in-process is the point: a retry is precisely what would create the duplicate.
2. **Run 2** — `/register` now answers `M_USER_IN_USE`. The recovery path logs in **with the same password** and adopts the account.

Step 2 only works because `password` is a declared, required input and therefore stable across invocations. That is the property the random fallback lacked, and why removing it in round 2 was the fix rather than persisting the generated value — the two findings are the same bug seen from opposite ends.

`TestBotAccountRecoveryIsStableAcrossRuns` now pins both halves in one test: run 1 must fail with exactly one register attempt, run 2 must recover using the declared password.

### Also in this push

`errors.As` across matrix, matching onepassword and attic. `isUserInUse` is the one that matters — it drives the recovery path above, so one added `%w` in httpx would turn recovery into a hard failure on an account that already exists. Pinned by `TestIsUserInUseSeesThroughWrapping`.

`gofmt` clean, `go vet` clean, `go test -race ./...` green across all 7 packages.

jmmaloney4:
## Round 5 (`8672bef`) — both fixed

Two more instances of the same drift class this review has been reliably finding: an input that reaches the server exactly once, and a `Diff` that doesn't watch it.

### `Room.Diff` ignored `accessToken` — fixed

Rotating the bot token produced no change to apply, so state kept the old value and every later `Read`, `Update` and `Delete` kept authenticating with a credential that no longer works. The trap is that supplying a fresh token *couldn't* recover it — `Diff` reported nothing to do, so the new value never reached state.

Now an in-place update, **never** a replacement: the token says nothing about the room, and recreating one would abandon the live room and its history. Same reasoning as the `transport` diff on `onepassword:Item`, which this had simply missed.

`TestRoomDiffCatchesTokenRotation` asserts the update fires, that **no** property forces replacement, that the new token lands in the returned state, and that zero homeserver calls are made (rotating a credential changes nothing server-side).

### `BotAccount.Create` didn't validate the response — fixed

It assigned `out.ID = res.UserID` straight from the response. A 2xx carrying no `user_id` or `access_token` would leave Pulumi tracking an account it cannot address while the registered user stays live on the homeserver. `Room.Create` already had the equivalent empty-`room_id` guard; this was an inconsistency, not a design choice.

The display-name call also moved *after* the check, so nothing acts on a response that is about to be rejected. `TestBotAccountCreateRejectsIncompleteResponse` covers missing `user_id`, missing `access_token`, and `{}`, and asserts the displayname request never fires.

### Note on the stale `/improve` entry

The `Password not persisted` suggestion under **Previous suggestions** is from `2e49f7f` and no longer applies — the random-password fallback it describes was removed in round 2, and `password` is now a required input. The current-HEAD suggestion list does not repeat it.

`gofmt` clean, `go vet` clean, `go test -race ./...` green across all 7 packages.

jmmaloney4:
## Round 6 (`563f451`) — you caught me making it worse

This is the best finding of the review, and it's a defect I introduced in round 3 rather than one inherited from the dynamic provider.

Matrix deactivation is **irreversible**: the user id is permanently retired and cannot be re-registered. So `DeleteBeforeReplace: true` meant a routine password rotation would deactivate `@alerts-bot`, burn the username forever, fail the follow-up `/register`, and fail the recovery login too (the account is deactivated) — stranding both the resource and the name with no way back.

I reasoned about the *collision* and never asked whether `Delete` was reversible. It isn't.

### `DeleteBeforeReplace` is now `false` unconditionally

Creating first isn't a *working* replacement when username and homeserver are unchanged — `/register` answers `M_USER_IN_USE` and recovery tries the new credential against an account holding the old one. But it fails **safely**: the apply errors and the live account is untouched.

Between a loud failure and an irreversible one, take the loud failure. A test now asserts `DeleteBeforeReplace == false` for *every* diff shape, with the reason in the failure message, so nobody re-derives my mistake.

### Password rotation gets a real path instead

That leaves rotation with no working replacement, so `Update` now calls Matrix's change-password API in place, authenticating the UIA stage with the old password from state. `password` moves from `UpdateReplace` back to `Update`. This removes the only routine trigger for a same-account replacement, which is what made the dangerous path reachable at all.

**`logout_devices` is explicitly `false`.** The API defaults it to `true`, which would invalidate the access token this resource stores *and* every consumer's session — a quieter disaster than the one being fixed. `TestBotAccountUpdateChangesPassword` asserts the flag, the new password, and that UIA authenticates with the **old** one.

On your `registrationToken` note: it stays a replacement, but is now safely ordered — it fails loudly rather than destructively. Whether rotating a homeserver registration token should recreate every bot at all is a real question, but it's the dynamic provider's existing semantics and changing it belongs in its own PR.

I did not verify tuwunel's specific deactivation semantics, and deliberately didn't make the fix depend on them: the safe ordering is correct whether or not tuwunel frees the username.

## Validation (local, per the flake-check gate)

| check | result |
|---|---|
| `provider-go-test` (gofmt + `go test -race ./...`, 7 pkgs) | **PASS** |
| `renovate-config` | **PASS** |
| `treefmt` | fails on **7 files, all inherited** — identical list on fresh `origin/main`, zero introduced by this PR |
| `tsc`, `vitest`, `pre-commit` | fail on a stale `pnpmDepsHash` (`sha256-wC91t…` declared vs `sha256-QkhFw…` actual), reproduced on fresh `origin/main`. This PR touches **zero TypeScript** — only `provider/matrix/*.go`, `provider/provider.go`, `flake.nix`. |

Baseline proof was done in a detached worktree at `origin/main`, comparing the failing file lists directly rather than assuming the failures were unrelated.

The stale `pnpmDepsHash` is worth its own issue — it blocks all local TS validation, and CI only stays green because it substitutes `node_modules` from Attic and never rebuilds it.

jmmaloney4:
## Round 7 (`4ab025e`) — ordering fixed

Correct, and it's a defect in the change-password code I added last round rather than anything inherited. Applied exactly as suggested.

`Update` ran the password change first and the display name second. Because a failed `Update` leaves Pulumi holding the **old** state, a display-name failure after a successful password change was unrecoverable: the next apply recomputes the same diff and retries, authenticating the UIA stage with `req.State.Password` — the old password — against an account that now has the new one. Every retry fails identically, forever. And as you note, `logout_devices: false` means the access token is still valid but cannot satisfy an `m.login.password` stage, so there's no way out from the client side.

The password change now goes **last**. Anything that fails before it leaves the credentials untouched and the whole `Update` cleanly retryable. Only a lost response from the change-password call itself can still strand the account — a far narrower window than "any later step failed" — and the error message now says so, so the operator knows state holds the previous password:

```
… if this call reached the homeserver, Pulumi state still holds the previous
password and the account must be reconciled server-side
```

Two tests pin the ordering from both directions: `TestBotAccountUpdateChangesPasswordLast` (happy path must issue displayname then password) and `TestBotAccountUpdateLeavesPasswordUntouchedOnEarlierFailure` (a failing display-name update must never reach the password call).

### On the `/improve` list

It is stale at `d674686` — its two ≥7 suggestions (`Create` response validation, `accessToken` in `Room.Diff`) both landed in `8672bef`. The third, adding a name guard to `Room.Create`, is already covered by `Room.Check`, which rejects an empty `name` before `Create` runs; a second guard would only fire if `Check` were bypassed entirely.

`gofmt` clean, `go vet` clean, `go test -race ./...` green across all 7 packages.

jmmaloney4:
## Merge-gate audit — `4ab025e`

Seven review rounds. Recording the gate explicitly rather than merging on a green-looking page.

### Review convergence

`## PR Reviewer Guide` is at the current HEAD (`4ab025e46868e448305a341e3dde96c08abe8d13`) and clean:

```
🔒 No security concerns identified
⚡ No major issues detected
```

`/improve` is **stuck at `563f451`** — it did not advance after six polls despite an explicit re-trigger. Its three suggestions are all already resolved, so nothing actionable is outstanding:

| suggestion | status |
|---|---|
| `Create` validates neither `UserID` nor `AccessToken` | implemented in `8672bef` |
| `AccessToken` omitted from `Room.Diff` | implemented in `8672bef` |
| add a name guard to `Room.Create` (importance 6) | already covered by `Room.Check`, which rejects an empty `name` before `Create` runs |

The last commit only reordered two calls in `Update` and added two tests, so it cannot have introduced a new finding of that shape.

### GitHub checks

`mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE`. The only two checks are `QUEUED` on self-hosted runners — `add-to-project` and `adr-management`. Neither builds or tests this code.

### Local validation (the real gate), at current HEAD

| check | result |
|---|---|
| `provider-go-test` — `gofmt -l` + `go test -race ./...` | **PASS**, 7 test packages |
| `renovate-config` | **PASS** |
| `treefmt` | fails on 7 files — **identical list on fresh `origin/main`**, `comm -13` empty |
| `tsc` / `vitest` / `pre-commit` | fail on a stale `pnpmDepsHash` (`sha256-wC91t…` declared vs `sha256-QkhFw…` actual), **reproduced on fresh `origin/main`** |

Baseline proof was done in a detached worktree at `origin/main` and compared by failing-file list, not assumed. This PR touches zero TypeScript — only `provider/matrix/*.go`, `provider/provider.go`, `flake.nix`.

### What the seven rounds actually found

Three of the last four findings were defects **introduced while fixing the previous round's finding**, each in a direction the prior round hadn't examined:

1. `Diff` ignored `password` / `isDirect` / cleared `displayName` → added them.
2. Adding `password` as a replacement exposed that `deleteBeforeReplace: false` was wrong for a same-account replace → set it true.
3. **That was worse.** Matrix deactivation is irreversible; delete-first would have burned the username on a routine password rotation → reverted to `false` unconditionally and gave rotation a real in-place path.
4. That change-password call ran before the display-name update, so a display-name failure stranded the account behind an unusable UIA credential → moved it last.

Every one is pinned by a test whose failure message states the reason, so the reasoning survives the next person who tries to "simplify" it.

### Follow-ups filed separately

- Stale `pnpmDepsHash` blocks all local TypeScript validation; CI only stays green because it substitutes `node_modules` from Attic and never rebuilds it.
- treefmt has no Go formatter wired up — the `provider-go-test` check now gates on `gofmt -l` to cover that gap.

Merging.

PR #348, opened 2026-08-03
fix(pulumi): stamp the plugin binary's real version via ldflags

jmmaloney4:
## Summary

The released `v0.20.2` plugin reports its own version to the Pulumi engine as the literal string `"dev"` instead of `"0.20.2"`. Found in production, not in testing: `pulumi preview -C deploy/services/litellm -s prod` (garden#1582, the just-merged plugin-wiring PR) failed outright with

```
error: Invalid character(s) found in major number "dev"
```

attributed to `pulumi:providers:sector7::default_0_20_2`, before any resource-level diffing could even start. This is **not litellm-specific** — every operation against every resource this provider owns fails the same way, since `"dev"` is not valid semver and the engine rejects it at the provider handshake, before `Check`/`Diff`/anything resource-specific ever runs.

## Root cause

`packages.pulumi-resource-sector7` in this repo's own `flake.nix` never passed `ldflags` to `buildGoModule`, so `provider/version/version.go`'s `var Version = "dev"` default was never overwritten — in every build, including the one already released.

`version.go`'s own comment explains why this slipped through: it said the injection happened "by the Nix derivation (`pkgs/pulumi-resource-sector7` in jackpkgs)" — the *original* plan, before packaging moved into this repo's own flake (a documented deviation, see the migration plan's STATUS section). The ldflags line never got carried over to the new location, and the stale comment pointed at the wrong place, so nothing flagged the gap during review.

## Fix

Adds `ldflags = ["-X" "…/version.Version=${providerVersion}"]` to the `buildGoModule` derivation, and corrects the stale comment in `version.go`.

## Regression guard: `checks.provider-version-stamped`

Deliberately **not** a `strings`/`grep` scan of the binary. I tried that first and it produced a false negative I had to chase down: Go's `-ldflags -X` stores the string with no guaranteed null-byte boundary before it, so a raw byte-run scanner can merge it with adjacent binary data — the correctly-injected value showed up as `"L0.20.2"`, an unrelated preceding byte glued onto the front. Worse, this *exact* version number coincidentally collides with an unrelated vendored dependency's own pinned version elsewhere in the binary (`go-openapi/jsonreference@v0.20.2`), so even a plain substring match would have given a **false pass** with ldflags completely broken.

Instead the check runs `pulumi package get-schema <bin>` and asserts the JSON `version` field — the same code path the Pulumi engine itself exercises, with no string-boundary ambiguity. Verified:
- reports literally `"dev"` against the pre-fix binary (reproduced the exact bug)
- reports the correct version after the fix
- **fails loudly** when the ldflags line is removed (negative control) — actually fails even earlier than my own assertion, at `get-schema`'s own schema-binding step, which refuses to parse a non-semver version

## Functional verification against the actual bug

With this fix applied via a local flake override (`--override-input sector7 path:…`, nothing merged yet), ran the real acceptance test: `pulumi preview -C deploy/services/litellm -s prod --show-sames`.

Every `sector7:litellm:ApiKey`/`Team`/`KeyRecord`/`TeamRecord` resource and every `RandomPassword` child (the 6 live API key secrets) shows `(same)` under its retyped URN — the alias absorbed the dynamic-provider-to-plugin retype with **zero** diff, exactly as designed:

```
sector7:litellm:ApiKey$sector7:litellm:KeyRecord::adr-compliance-bot-key-key: (same)
  [provider: …pulumi-nodejs::default::93dfeeb9… => …sector7::default_0_20_2::[unknown]]
```

The only pending changes in that preview (8 update, 6 replace) are unrelated pre-existing drift in `codex-proxy`/quota-exporter Nix images — confirmed by URN, not by absence, that none of them touch `sector7:litellm:` or `randomPassword`.

## Test plan

- [x] `gofmt -l provider/` clean
- [x] `go build`, `go vet`, `go test -race ./...` — all 7 packages pass
- [x] `nix build .#checks.x86_64-linux.provider-version-stamped` — pass, and fails correctly on a negative control
- [x] `nix build .#checks.x86_64-linux.treefmt` — fails only on 7 pre-existing files unrelated to this diff (sector7#347)
- [x] End-to-end functional verification against real litellm/prod state, as above

## Not in this PR

Cutting the release and re-wiring garden — that's the next step once this merges.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -61,6 +61,14 @@
         ];
 
         renovateConfigPaths = map (path: "${self.outPath}/${path}") renovateConfigFiles;
+
+        # Single source of truth for packages.pulumi-resource-sector7 AND
+        # checks.provider-version-stamped, so the two can never desync — the
+        # exact failure mode (a build path deriving the version one way, a
+        # check asserting it a different way) that let v0.20.2 ship reporting
+        # itself as "dev".
+        providerVersion =
+          (builtins.fromJSON (builtins.readFile ./packages/sector7/package.json)).version;
       in {
         jackpkgs.just.cut = {
           enable = true;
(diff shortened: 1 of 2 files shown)


jmmaloney4:
## Both /improve suggestions applied (`c432f73`)

**Deduplicate version parsing** — correct, and worth calling out precisely: this was the same *shape* of bug this PR exists to fix, one derivation deriving the version one way and another asserting it a different way. Hoisted `providerVersion` to a single `let`-binding shared by both `packages.pulumi-resource-sector7` and `checks.provider-version-stamped`.

**Escape shell interpolation** — agreed, applied via the env-var route rather than `lib.escapeShellArg`: `want = providerVersion;` in the derivation's attrset, referenced as `$want` in the script. Nix passes derivation env values through the environment, not through string interpolation into the script text, so nothing about the value's content is ever handed to the shell for interpretation — stronger than escaping a specific set of characters.

Re-verified after the refactor:
- `gofmt`/`go build`/`go vet`/`go test -race ./...` — clean, all 7 packages
- `nix build .#checks.x86_64-linux.provider-version-stamped` — pass
- Negative control (remove `ldflags`) — still fails correctly, at the same schema-binding step
- `nix run nixpkgs#alejandra -- --check flake.nix` — clean

jmmaloney4:
## Merge-gate audit

- Reviewer guide clean at HEAD (`c432f730`): no security concerns, no major issues.
- `/improve` clean at HEAD: "No code suggestions found for the PR" (posted after both rounds of fixes were pushed).
- `mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE` — the only non-passing check is `adr-management`, stuck `QUEUED` on the same zero-runner self-hosted fleet that's affected every PR tonight. Nothing is actually failing.
- Local: `go build`/`go vet`/`go test -race ./...` clean (7 packages), `provider-version-stamped` passes and correctly fails on a negative control, `alejandra --check` clean.
- Functionally verified against the actual production bug: with this fix applied via a local override, `pulumi preview` against `litellm/prod` completes cleanly, with every litellm resource and RandomPassword child showing `(same)` under its retyped URN.

Merging.

PR #349, opened 2026-08-03
fix(provider): wire a real kube.Transport into port-forwarded resources

jmmaloney4:
## Summary

Four resources — `litellm.TeamRecord`, `litellm.KeyRecord`, `onepassword.Item`, `attic.Cache` — were registered as bare struct literals, leaving their `kube.Transport` **interface field nil**. The first operation to actually open a port-forward segfaulted the whole provider process:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18]
  ...provider/litellm.connect(...)        client.go:43
  ...provider/litellm.KeyRecord.Read(...) key.go:315
```

Found by running `pulumi refresh` against `litellm/prod` right after the migration landed — the acceptance test this entire migration exists to make possible.

`internal/kube/kube.go`'s own package comment already said *"production wires SPDYTransport"*. It simply never did.

## Why this survived five releases

Nothing in the test or deploy path ever exercised the zero value:

- **Every unit test** constructs these types with an explicit `kube.Fake`, so the nil path is unreachable in `go test`.
- **`Check` and `Diff` never touch the transport**, so `pulumi preview` and `pulumi up` stayed green for as long as state was byte-identical — which is *exactly* what the alias migration was engineered to produce. The migration succeeding is what kept the bug hidden.

A nil interface field is invisible to the compiler and to any test that dutifully populates it.

## Fix

One shared `*kube.SPDYTransport` passed to the four resources that need it. Sharing is deliberate: it caches `*rest.Config` per kubeconfig payload behind a mutex, so a single instance parses each kubeconfig once per process, while still opening a fresh forward per CRUD call.

`attic.Token` (mints a JWT locally), `d1`, `r2` and `matrix` (public APIs) correctly have no transport field at all — I verified this by scanning every resource struct rather than assuming.

## Regression guard

`TestEveryKubeBackedResourceHasATransport` is a **structural** check, because the bug is structural and nothing behavioural caught it.

The resource list moves into `resources(tr)` so the test can reflect over *exactly what `New()` registers* rather than a reconstruction — a reconstruction would be worthless here, since the defect was precisely that the real list and a correct-looking one had diverged. It digs out `derivedResourceController.receiver` by reflection, and guards itself with an expected-count assertion so a `pulumi-go-provider` refactor makes it fail loudly instead of passing vacuously.

Verified in both directions:
- passes on this tree
- reverting one resource to a bare literal fails it with `litellm.KeyRecord.Transport is nil: this segfaults the entire provider process...`

## Verified against the real bug

With this build linked, `pulumi refresh -C deploy/services/litellm -s prod`:

- **zero** `panic` / `SIGSEGV` occurrences (was: process death)
- the provider now loads the kubeconfig, authenticates, and issues a real Kubernetes API call
- it surfaces a clean, actionable error instead of a segfault

## Separate pre-existing issue this now makes visible

With the crash gone, `refresh` reveals what was underneath:

```
sector7: reading deployment litellm/litellm: deployments.apps "litellm" is forbidden:
User "pulumi-zeus" cannot get resource "deployments" in API group "apps" in the namespace "litellm"
```

`kubectl auth can-i` under the only available context confirms `pulumi-zeus` has **none** of the three permissions a port-forward requires:

| permission | result |
|---|---|
| `get deployments -n litellm` | no |
| `create pods/portforward -n litellm` | no |
| `list pods -n litellm` | no |

This is **not** caused by this PR or by the migration — the TypeScript dynamic provider needed the exact same permissions and used the same ambient kubeconfig. It was masked by two successive earlier failures: first `ERR_MODULE_NOT_FOUND` (the serialised-closure bug that motivated this whole migration), then this segfault. It is an RBAC/infrastructure matter, tracked separately — deliberately not addressed here.

Practical impact: `preview`/`up` on unchanged resources are unaffected (confirmed — litellm/prod's migration `up` succeeded earlier). `refresh`, and any real change to a LiteLLM key/team, Attic cache, or 1Password item, needs the RBAC granted.

## Test plan

- [x] `gofmt -l`, `go build`, `go vet` clean
- [x] `go test -race ./...` — all 8 packages pass
- [x] Negative control on the new test — fails correctly
- [x] Real `pulumi refresh` against `litellm/prod` — panic gone

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/jmmaloney4/sector7/provider/attic"
 	"github.com/jmmaloney4/sector7/provider/d1"
+	"github.com/jmmaloney4/sector7/provider/internal/kube"
 	"github.com/jmmaloney4/sector7/provider/litellm"
 	"github.com/jmmaloney4/sector7/provider/matrix"
 	"github.com/jmmaloney4/sector7/provider/onepassword"
@@ -36,6 +37,20 @@ const Name = "sector7"
 // with the dynamic provider it replaces, so that retyping a live resource via
 // `aliases: [{ type: "pulumi-nodejs:dynamic:Resource" }]` diffs to nothing.
 func New() (p.Provider, error) {
+	// ONE transport shared by every port-forwarded resource.
+	//
+	// Sharing is deliberate: SPDYTransport caches *rest.Config per kubeconfig
+	// payload, so a single instance parses each kubeconfig once for the whole
+	// process, and its mutex makes that cache safe under Pulumi's concurrent
+	// gRPC serving. It caches configs, never forwards — each CRUD call still
+	// opens its own port-forward.
+	//
+	// Passing it is NOT optional. kube.Transport is an interface, so a resource
+	// registered as a bare `litellm.KeyRecord{}` carries a nil transport and
+	// SEGFAULTS the whole provider process the moment it reaches connect() —
+	// see resources() below.
+	transport := &kube.SPDYTransport{}
+
 	return infer.NewProviderBuilder().
 		WithNamespace("jmmaloney4").
 		WithDisplayName("sector7").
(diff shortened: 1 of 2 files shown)


jmmaloney4:
## Merge-gate audit

- Reviewer guide clean: no security concerns, no major issues.
- `/improve` clean: "No code suggestions found for the PR." Both gates landed after open with no commits pushed since, so both reflect current HEAD (`39e49445`).
- `mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE` — only `add-to-project` and `adr-management` queued on the self-hosted fleet, which has had **zero registered runners** all session. Nothing failing.
- Local: `gofmt`/`go build`/`go vet` clean, `go test -race ./...` green across all 8 packages, `nix build .#checks.x86_64-linux.provider-go-test` pass.
- New test verified in **both** directions: passes on this tree, and fails with `litellm.KeyRecord.Transport is nil` when a single resource is reverted to a bare literal.
- Verified against the real production bug: `pulumi refresh` against litellm/prod no longer panics (zero SIGSEGV).

Follow-up filed as jmmaloney4/garden#1585 for the RBAC gap this exposes — pre-existing, infrastructure, deliberately out of scope here.

Merging.

Issue #347, opened 2026-08-02
build: treefmt check is red on main (7 unformatted files)


jmmaloney4:
## Problem

`nix build .#checks.x86_64-linux.treefmt` fails on `origin/main` (verified at `0a241b8` in a fresh detached worktree). Seven committed files are not in the formatter's canonical form:

```
docs/renovate/README.md
docs/renovate/workflows.md
packages/sector7/pulumi-config/index.ts
packages/sector7/pulumi-config/op-secret-helpers.ts
packages/sector7/tests/op-secret-helpers.test.ts
packages/sector7/tests/port-forward.test.ts
renovate/minor-patch-automerge.json
```

## Why it matters

A permanently-red formatter check is worse than no check: every PR inherits the failure, so `treefmt` carries no signal about whether *that PR* introduced drift. Establishing that for #344 meant building `treefmt` on a fresh `origin/main` worktree and diffing the two failing-file lists by hand — which is not a thing anyone will do routinely.

## Likely cause

A formatter version bump in the jackpkgs pin changing canonical output (the `renovate/minor-patch-automerge.json` diff is a JSON array being expanded to one element per line, which reads like a `prettier`/`biome` behaviour change rather than a hand edit).

That would also explain why CI is green: CI substitutes the check result from Attic and never re-runs it against the current formatter.

## Fix

Run `nix fmt` and commit the result. Worth confirming the diff is purely mechanical first — if a formatter version changed, the reformat should land as its own commit with the pin bump referenced, not folded into unrelated work.

## Related

- #346 — treefmt has no Go formatter at all, a separate gap in the same config
- #345 — same "CI is green because it substitutes from Attic and never rebuilds" failure mode, applied to `pnpmDepsHash`

Issue #345, opened 2026-08-02
build: stale pnpmDepsHash breaks all local TypeScript validation


jmmaloney4:
## Symptom

`nix flake check` cannot build `tsc`, `vitest` or `pre-commit` on a machine without a warm cache:

```
error: hash mismatch in fixed-output derivation '…-pnpm-pnpm-deps.drv':
         specified: sha256-wC91tGjpyAS929g21f7tfXN8Enm61aYSERlLhZ+3qP8=
            got:    sha256-QkhFw4knCylOpGSpYoCd2Na14c/eeXM+q8J/id5fo+U=
```

`jackpkgs.nodejs.pnpmDepsHash` in `flake.nix` no longer matches what `pnpm fetch` actually produces for the committed `pnpm-lock.yaml`.

## Why it has stayed invisible

CI substitutes the prebuilt `node_modules` store path from Attic and never rebuilds it, so the fixed-output hash is never checked there. Every CI run is green. The failure only appears for someone whose cache is cold — which is exactly the person least equipped to tell it apart from a mistake of their own.

Reproduced on a fresh detached worktree at `origin/main` (commit `0a241b8`), so it is not branch-specific.

## Impact

`nix develop` also fails, since the devshell depends on the same derivation. That makes the local flake-check gate unusable for the whole TypeScript half of the repo — checks that CI *appears* to be covering.

## Suggested fix

Recompute and commit the hash. Worth understanding *why* it drifted before updating it: if the lockfile and the cached artifact genuinely disagree, the cached `node_modules` CI has been reusing is stale relative to `pnpm-lock.yaml`, which is a bigger problem than the hash line.

A CI job that builds `node_modules` **without** substituters, even occasionally, would stop this recurring — the current setup is structurally incapable of noticing.

## Context

Found while running `nix flake check` as the merge gate for #344 (the Pulumi provider plugin migration). Not caused by that PR, which touches zero TypeScript.

PR #350, opened 2026-08-03
feat(litellm): accept an explicit kubeconfig for the admin port-forward

jmmaloney4:
## Summary

`LiteLLMApiKey` and `LiteLLMTeam` open their admin port-forward using the **ambient** kubeconfig, while everything else in garden's litellm stack uses the **platform** kubeconfig — its `k8s.Provider` is bound to it via `getK8sProvider`, and `OnePasswordItem` *in that same file* already passes it explicitly. The LiteLLM resources were the only ones left reading whatever `kubectl` happens to point at.

On a machine whose kubeconfig is the narrower `pulumi-zeus` deploy identity, every operation is refused:

```
deployments.apps "litellm" is forbidden: User "pulumi-zeus" cannot get
resource "deployments" in API group "apps" in the namespace "litellm"
```

`deploy/lib/index.ts` already documents explicit kubeconfig as the intended pattern for *"dynamic resources that open their own in-process Kubernetes port-forward and therefore need the raw kubeconfig rather than a bound `k8s.Provider`"*. This closes the gap, rather than widening a deploy identity's RBAC to match an identity the stack already has.

## Design notes

**Optional, not required** — resources retyped by the plugin migration keep decoding; no live state carries this field.

**Secret** — a kubeconfig holds client certificates and keys. Matches `onepassword.ItemArgs.Kubeconfig`.

**Passed through raw in TypeScript, not defaulted to `""`** — `undefined` is omitted from the inputs entirely, so a caller that sets no kubeconfig keeps decoding to the Go zero value and diffs to nothing. Same reasoning as the `maxBudget` sentinel fix.

**Both `Diff`s now route through `AdminTarget.Changed`** instead of hand-listing fields, so `KeyRecord` and `TeamRecord` cannot disagree about what an admin-target change is. Adding a field to one and forgetting the other would strand a rotated kubeconfig in one resource's state — the same class of bug as the four resources that shipped with a nil transport.

**`kube.Fake` gains `LastTarget`.** Without it a resource can silently drop part of its connection inputs and every test still passes, because the Fake ignores the `Target` and returns the same `BaseURL` regardless. The failure is visible only in production, as a port-forward opened against the wrong identity. This is the same reason the nil-transport bug (#349) survived five releases.

## Verification

Both directions, as usual:

- `TestAdminTargetKubeconfigReachesTheTransport` passes — and removing the single line that forwards `Kubeconfig` fails it with `kubeconfig did not reach kube.Target: got ""`
- `TestKubeconfigChangeIsAnAdminTargetDiff` pins that a rotation is an **in-place update, never a replace** of a live key, and that unchanged inputs still diff to nothing (what keeps the six live credentials from churning)

## Test plan

- [x] `nix build .#checks.x86_64-linux.provider-go-test` — pass (gofmt + `go test -race ./...`, all packages)
- [x] Negative control on the new propagation test — fails correctly
- [x] `tsc -p tsconfig.build.json --noEmit` — clean (run directly; see below)
- [x] `vitest run` — 304 tests pass
- [x] `alejandra --check flake.nix` — clean

Two checks fail for reasons **proven pre-existing**, both filed earlier:

| check | cause | proof |
|---|---|---|
| `tsc`, `vitest` (nix) | stale `pnpmDepsHash` (#345) | byte-identical hash mismatch (`wC91t…` vs `QkhFw…`) |
| `treefmt` | 7 unformatted files (#347) | none of them mine — flagged list is `docs/renovate/*`, `pulumi-config/*`, two tests, one renovate json |
| `vitest` suite `renovate-pulumi.test.ts` | "expected exactly one Pulumi customManager, found 2" | reproduced identically on a clean `origin/main` worktree: `1 failed, 24 passed, 304 tests` |

Because #345 blocks the nix `tsc`/`vitest` checks entirely, I compiled and tested the TypeScript directly inside `.#ci-pnpm` instead.

## Scope

`attic.Cache` has the **identical** gap — no `Kubeconfig` field, `connect` never passes one — and is deliberately untouched. `attic/prod` is a separate stack with 13 live resources including host tokens, and nothing is blocked on it today. Worth doing when attic's TypeScript is retyped.

### DIFF
--- a/packages/sector7/litellm/admin.ts
+++ b/packages/sector7/litellm/admin.ts
@@ -61,6 +61,11 @@ function pluginOpts(
  */
 function teamProviderInputs(args: LiteLLMTeamArgs) {
 	return {
+		// Passed through rather than defaulted to "": undefined is omitted from
+		// the inputs entirely, so a resource that never set a kubeconfig keeps
+		// decoding to the Go zero value and diffs to nothing. Same reasoning as
+		// `maxBudget` below.
+		kubeconfig: args.kubeconfig,
 		proxyNamespace: args.proxyNamespace,
 		masterKey: args.masterKey,
 		proxyDeploymentName: pulumi.output(args.proxyDeploymentName ?? "litellm"),
@@ -130,6 +135,11 @@ function keyProviderInputs(
 	keyValue: pulumi.Output<string>,
 ) {
 	return {
+		// Passed through rather than defaulted to "": undefined is omitted from
+		// the inputs entirely, so a resource that never set a kubeconfig keeps
+		// decoding to the Go zero value and diffs to nothing. Same reasoning as
+		// `maxBudget` below.
+		kubeconfig: args.kubeconfig,
 		proxyNamespace: args.proxyNamespace,
 		masterKey: args.masterKey,
 		proxyDeploymentName: pulumi.output(args.proxyDeploymentName ?? "litellm"),
--- a/packages/sector7/litellm/config-types.ts
+++ b/packages/sector7/litellm/config-types.ts
@@ -237,6 +237,24 @@ export interface LiteLLMGeneratedConfig {
 }
 
 export interface LiteLLMAdminTargetArgs {
+	/**
+	 * Kubeconfig YAML used to open the admin port-forward. Omit to use the
+	 * ambient default config.
+	 *
+	 * Pass this. The ambient config is whatever `kubectl` happens to point at,
+	 * which is usually NOT the identity the rest of the stack deploys with:
+	 * garden binds its `k8s.Provider` to the platform kubeconfig via
+	 * `getK8sProvider`, and `OnePasswordItem` already takes this same input for
+	 * exactly this reason. Leaving it unset is why a narrower ambient identity
+	 * gets refused:
+	 *
+	 *   deployments.apps "litellm" is forbidden: User "pulumi-zeus" cannot get
+	 *   resource "deployments" in API group "apps" in the namespace "litellm"
+	 *
+	 * Optional rather than required so resources retyped by the plugin
+	 * migration keep decoding — no existing state carries it.
+	 */
+	kubeconfig?: pulumi.Input<string>;
 	proxyNamespace: pulumi.Input<string>;
 	masterKey: pulumi.Input<string>;
 	proxyDeploymentName?: pulumi.Input<string>;
--- a/provider/internal/kube/kube.go
+++ b/provider/internal/kube/kube.go
@@ -56,10 +56,19 @@ type Fake struct {
 	// Connects counts calls, so tests can assert the per-operation forward
 	// behaviour (each CRUD method opens its own).
 	Connects int
+	// LastTarget is the most recent Target received.
+	//
+	// Without it a resource can silently drop part of its connection inputs —
+	// Kubeconfig especially — and every test still passes, because the Fake
+	// ignores the Target and hands back the same BaseURL regardless. The only
+	// visible symptom is in production, as a port-forward opened against the
+	// wrong cluster identity.
+	LastTarget Target
 }
 
 func (f *Fake) Connect(_ context.Context, t Target) (*Conn, error) {
 	f.Connects++
+	f.LastTarget = t
 	host := f.HostVal
 	if host == "" {
 		host = t.Deployment + "." + t.Namespace + ".svc.cluster.local"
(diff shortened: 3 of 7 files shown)


jmmaloney4:
## Merge-gate audit

Head at audit time: `417cdb243017ca45de8c79e791b17aa22860e31d` (single commit, nothing pushed after PR open).

### Review status — converged

| Gate | Result |
| --- | --- |
| `/review` — `## PR Reviewer Guide` | Clean. `No major issues detected`, `No security concerns identified`, no focus areas. |
| `/improve` — `## PR Code Suggestions` | Clean. `No code suggestions found for the PR.` |

Both gate comments landed after the sole commit (commit `16:30:43Z`; `/review` `16:32:05Z`; `/improve` `16:33:40Z`) and no commit was pushed afterwards, so both describe the current head. Neither carries a `<!-- <sha> -->` marker, which is the documented clean-first-run shape; all four fallback conditions hold — both comments posted after open, no post-open commits, both clean, local validation green on this head.

Nothing was triaged away as a false positive because nothing actionable was raised.

### Independent review (the bot is not the only gate)

Checked the four things a new secret provider input can plausibly get wrong here:

- **Struct tag / secrecy** — `Kubeconfig string \`pulumi:"kubeconfig,optional" provider:"secret"\`` is byte-identical to the `onepassword.ItemArgs.Kubeconfig` precedent it cites. Secret marking is correct for a credential-bearing YAML blob.
- **Schema regeneration** — no checked-in `schema.json`/SDK tree in this repo; the Pulumi schema is derived at runtime by `infer` from the Go structs, so adding a field needs no generated-file refresh.
- **State compatibility** — `migrations.go` embeds `AdminTarget` in the historical v0 sentinel shapes. The new field is `optional`, so old serialized state decodes it to `""` and the migrations copy it through unchanged. No live resource gains a spurious diff.
- **`AdminTarget.Changed`** — struct `!=` is safe (every field is comparable) and is now the single source of truth for both `KeyRecord.Diff` and `TeamRecord.Diff`, which is what stops the two from disagreeing about an admin-target change. `TestKubeconfigChangeIsAnAdminTargetDiff` pins both the in-place-update behavior and the no-replace invariant.

The `kube.Fake.LastTarget` addition is the load-bearing test-infra piece: before it, dropping any connection input was undetectable, since the Fake returned the same `BaseURL` regardless of `Target`.

### CI — queued, will not run

All 3 GitHub checks are `QUEUED`. The self-hosted fleet has 3 registered runners and **all 3 report `status: offline`**, so nothing will ever pick these up:

```
jmmaloney4-sector7-6kwsg-runner-45mjt  offline  busy=false
jmmaloney4-sector7-6kwsg-runner-blvb8  offline  busy=false
jmmaloney4-sector7-6kwsg-runner-ff7nm  offline  busy=false
```

`mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE` — the `UNSTABLE` is entirely the queued-runner artifact. Local validation is the gate.

### Local validation — green on this head

| Command | Result |
| --- | --- |
| `nix build .#checks.x86_64-linux.provider-go-test --no-link` | **pass** (exit 0) — gofmt + `go test -race ./...`, covers every Go file in the diff |
| `nix develop .#ci-pnpm --command … pnpm exec tsc -p tsconfig.build.json --noEmit` | **pass** (exit 0) — covers `admin.ts` + `config-types.ts` |

### Inherited failures — pre-existing, not introduced here

- **`tsc` / `vitest` nix checks** fail on a stale `pnpmDepsHash` (`specified: sha256-wC91t…` vs `got: sha256-QkhFw…`). Tracked as #345. Worked around by running `tsc` directly in the `ci-pnpm` devshell, which passes (above).
- **`treefmt` check** is red. Tracked as #347. I extracted the exact unformatted set from the build log and confirmed **none** of the 7 files is touched by this PR:
  `docs/renovate/README.md`, `docs/renovate/workflows.md`, `packages/sector7/pulumi-config/index.ts`, `packages/sector7/pulumi-config/op-secret-helpers.ts`, `packages/sector7/tests/op-secret-helpers.test.ts`, `packages/sector7/tests/port-forward.test.ts`, `renovate/minor-patch-automerge.json`.
  This PR's Go files are separately gofmt-verified by `provider-go-test`.
- **`tests/renovate-pulumi.test.ts`** fails with `expected exactly one Pulumi customManager, found 2`, reproduced identically on a clean `origin/main` worktree.

### Decision

Review converged with zero actionable findings, PR-scoped local validation green on the current head, and every red gate proven inherited. Merging with a merge commit (squash and rebase are disabled on this repo).

PR #351, opened 2026-08-03
fix(litellm): never replace a live key on an unknown upstream input

jmmaloney4:
## Summary

`KeyRecord.Diff` replaced on **any** `teamId` or `keyValue` difference — including a difference against the Go zero value. An input that is unknown at preview time (because the resource producing it is itself updating) decodes as `""` here, not as an unknown sentinel. So a key whose `teamId` comes from a `LiteLLMTeam` **output** compared `"personal"` against `""` and planned a replacement:

```
~ teamId : "personal" => [unknown]
+-sector7:litellm:KeyRecord: (replace)
```

That means **any** update to a team threatens to replace every one of its keys — six live production credentials in `litellm/prod` — on something as ordinary as changing a team's model list.

## How it was found

While previewing an unrelated change (wiring an explicit kubeconfig, #350), the migration runbook's gate fired. It names a `KeyRecord` replace as an explicit *stop and fall back to state surgery* condition, so the rollout halted there rather than proceeding on the assumption that the replace would evaporate once the team's real `teamId` resolved.

Pre-existing, not introduced by that change: no team had been updated since the plugin migration, so nothing had exercised this path.

## The asymmetry

`TeamRecord.Diff` has **always** carried this guard:

```go
if news.DesiredTeamID != "" && news.DesiredTeamID != managed {
```

The key side never did. Both replace triggers are now guarded on a non-empty new value.

## Deliberate tradeoff

Clearing `teamId` or `keyValue` back to empty no longer replaces. Neither is a real operation — a LiteLLM key always belongs to a team, and `Check` rejects an empty `keyValue` — whereas an unknown upstream is routine. Documented inline.

## Tests

`TestUnknownUpstreamInputNeverReplacesALiveKey` covers unknown `teamId`, unknown `keyValue`, and both at once. It separately asserts a **real** change to either still replaces, so the guard cannot silently blunt a genuine identity change.

Verified both directions:
- passes on this tree
- removing the `teamId` guard fails it with `an unknown upstream input must never replace a live key (teamId planned a replace); this is six production credentials`

## Test plan

- [x] `nix build .#checks.x86_64-linux.provider-go-test` — pass (gofmt + `go test -race ./...`)
- [x] Negative control — fails correctly

Known-inherited failures, unchanged and not chased: #345 (stale `pnpmDepsHash` breaks the `tsc`/`vitest` nix checks), #347 (treefmt red on 7 files, none in this diff).

## Follow-up

Once released, garden can finally wire `kubeconfig: getPlatformKubeconfig({ environment })` into the LiteLLM resources with a clean preview — the change that surfaced this.

### DIFF
--- a/provider/litellm/key.go
+++ b/provider/litellm/key.go
@@ -90,10 +90,31 @@ func (KeyRecord) Diff(_ context.Context, req infer.DiffRequest[KeyArgs, KeyState
 	olds, news := req.State, req.Inputs
 	diffs := map[string]p.PropertyDiff{}
 
-	if olds.KeyValue != news.KeyValue {
+	// Both replace triggers are guarded on a NON-EMPTY new value.
+	//
+	// An input that is unknown at preview time — because the resource
+	// producing it is itself being updated — decodes to the Go zero value
+	// here, not to an unknown sentinel. Without the guard, a key whose teamId
+	// comes from a LiteLLMTeam output sees "personal" != "" and reports
+	// UpdateReplace, so ANY update to a team threatens to replace every one of
+	// its keys:
+	//
+	//	~ teamId : "personal" => [unknown]
+	//	+-sector7:litellm:KeyRecord: (replace)
+	//
+	// That is six live production credentials in litellm/prod, triggered by
+	// something as ordinary as changing a team's model list. TeamRecord.Diff
+	// has always carried this guard (`news.DesiredTeamID != "" && …`); the key
+	// side never did, and nothing exercised it until a team actually changed.
+	//
+	// The tradeoff is deliberate: deliberately clearing teamId or keyValue back
+	// to empty no longer replaces. Neither is a real operation — a LiteLLM key
+	// always belongs to a team, and Check rejects an empty keyValue — whereas
+	// an unknown upstream is routine.
+	if news.KeyValue != "" && olds.KeyValue != news.KeyValue {
 		diffs["keyValue"] = p.PropertyDiff{Kind: p.UpdateReplace}
 	}
-	if olds.TeamID != news.TeamID {
+	if news.TeamID != "" && olds.TeamID != news.TeamID {
 		diffs["teamId"] = p.PropertyDiff{Kind: p.UpdateReplace}
 	}
 
--- a/provider/litellm/key_test.go
+++ b/provider/litellm/key_test.go
@@ -498,3 +498,61 @@ func TestKubeconfigChangeIsAnAdminTargetDiff(t *testing.T) {
 		t.Fatalf("unchanged inputs must not diff; got %+v", r.DetailedDiff)
 	}
 }
+
+// An unknown upstream input must never replace a live key.
+//
+// This is the exact shape that stopped the litellm/prod kubeconfig rollout:
+// garden's keys take `teamId: personalTeam.teamId`, an OUTPUT of the team. When
+// the team itself updates — for any reason — its outputs are unknown during
+// preview, and an unknown decodes here as the zero value rather than as an
+// unknown sentinel. Before the non-empty guard, all six live production keys
+// planned as replaces:
+//
+//	~ teamId : "personal" => [unknown]
+//	+-sector7:litellm:KeyRecord: (replace)
+//
+// A key is only ever replaced for a REAL, KNOWN change of identity.
+func TestUnknownUpstreamInputNeverReplacesALiveKey(t *testing.T) {
+	old := KeyState{KeyArgs: baseKeyArgs()}
+
+	for _, tc := range []struct {
+		name   string
+		mutate func(*KeyArgs)
+	}{
+		// "" is how infer surfaces an input that is unknown at preview time.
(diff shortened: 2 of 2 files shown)


jmmaloney4:
## Merge-gate audit

**Head:** `21e82b7b1d497c3eccdbad28427de970b2b97185` (single commit; nothing pushed after the PR was opened).

### Review convergence — clean, zero actionable findings

| Gate | Result |
| --- | --- |
| `/review` (`## PR Reviewer Guide`, auto) | No security concerns, no major issues detected, no recommended focus areas |
| `/improve` (auto, at open) | `No code suggestions found for the PR.` |
| `/improve` (manual re-trigger at current HEAD) | `No code suggestions found for the PR.` |

Neither auto-run comment carries a `<!-- sha -->` marker, so recency alone was not trusted. Convergence was established structurally instead: the branch has exactly one commit (`21e82b7`, 17:03:40Z), the PR opened at 17:04:04Z, and both gate comments landed after that (17:04:40Z / 17:04:55Z) with no subsequent push. The manual `/improve` re-run at 17:09:03Z then re-confirmed a clean result against the current HEAD, satisfying the normal-convergence rule directly rather than by inference.

PR-Agent raised no design or correctness objection.

### CI: not a gate here (offline runners)

All three self-hosted runners are registered but `offline`:

```
jmmaloney4-sector7-6kwsg-runner-45mjt  offline
jmmaloney4-sector7-6kwsg-runner-blvb8  offline
jmmaloney4-sector7-6kwsg-runner-ff7nm  offline
```

Consequently all three GitHub checks sit `QUEUED` and `mergeStateStatus` is `UNSTABLE` while `mergeable` is `MERGEABLE`. No check has failed — none has started. Local validation is the gate.

### Local validation — green

```
nix build .#checks.x86_64-linux.provider-go-test --no-link   # exit 0
```

That derivation (`flake.nix:181-205`) runs a `gofmt -l` drift check and then `go test -race ./...`. Both passed; unpiped exit code captured directly. Relevant package:

```
ok  github.com/jmmaloney4/sector7/provider/litellm  2.898s
```

All other provider packages `ok`, no `FAIL`, no gofmt drift.

### Independent verification of the deliberate tradeoff

This PR knowingly trades away one behaviour: clearing `teamId` or `keyValue` to empty no longer forces a replace. The safety argument rests on two claims, both verified against source rather than taken on trust.

**1. Unknown really does decode to `""`.** Confirmed at the library level, not just asserted. In pulumi-go-provider v1.4.1, `infer/internal/ende/ende.go:183`:

```go
case v.IsComputed():
    el := e.walk(v.Input().Element, path, typ, true)
    e.mark(change{path: path, computed: true})
    return el
```

The decoder walks the *element* of a computed value and records the unknown-ness only in the encoder's change set — so the typed struct field receives the Go zero value. The premise is exact.

**2. The guard is the only mechanism available.** `infer.DiffRequest[I, O]` (`infer/resource.go:160-167`) carries only the typed `State O` and `Inputs I` — there is no raw `property.Map` on it, so a typed `Diff` structurally *cannot* distinguish unknown from empty. (By contrast `CheckRequest` does carry `property.Map`.) The non-empty guard is therefore a forced choice, not a shortcut past a cleaner API.

**3. `Check` really does reject an empty `keyValue` — claim holds.** `provider/litellm/key.go:69-79` puts `keyValue` in the required-non-empty loop and emits `CheckFailure{Property: "keyValue", Reason: "keyValue is required"}`. `Check` runs ahead of `Diff` on every operation, so an explicitly-cleared `keyValue` can never reach the weakened trigger. The tradeoff is sound as written.

Two corollaries worth recording, neither blocking:

- Because `Check` decodes unknowns to `""` by the same mechanism as (1), an *unknown* `keyValue` also fails `Check` — before `Diff` is ever consulted. So the `keyValue` arm of the new guard is defence-in-depth rather than the live path; the `teamId` arm is the one that actually fixes prod, since `teamId` is `optional` and deliberately absent from `Check`'s required list. This asymmetry is pre-existing and untouched by this PR.
- `Update` persists `KeyState{KeyArgs: req.Inputs}` wholesale. If an operator genuinely cleared `teamId` *and* changed another field in the same update, state would record `teamId: ""` while the remote key kept its team, and `findKeyHashesByAlias` (which matches on alias **and** team) would not adopt or delete that key on a later replace. This is the narrow, operator-initiated tail of the accepted tradeoff — it cannot fire from an unknown upstream, because `Update` only ever sees resolved inputs during apply. Recorded as accepted, not fixed here.

### Inherited failures — not introduced by this PR, not fixed here

- **#345** — stale `pnpmDepsHash` reds the `tsc` / `vitest` checks. This PR touches no TypeScript.
- **#347** — `treefmt` red on 7 files on `main`, none of them in this diff.

Both remain open and are repo-wide debt; per the standing rule they are cleaned up in separate mechanical PRs, not absorbed as a drive-by here.

### Gate summary

Review converged clean at current HEAD; PR-scoped local validation green; the one deliberate behavioural tradeoff independently verified against provider and library source; the only red checks are inherited and tracked. Merging with a merge commit (squash and rebase are disabled on this repo).

PR #353, opened 2026-08-03
feat(d1): migrate __provider out of dynamic-provider state

jmmaloney4:
## Summary

First slice of the next migration family: **uptime/prod's `D1Query`**, chosen as the lowest-blast-radius live dynamic resource remaining.

Live state written by the dynamic provider carries `__provider` — the serialized JavaScript closure this whole plugin exists to stop storing. No struct declares it, and `infer` treats an unrecognized property as a decode **failure**, not something quietly ignored:

```
Unrecognized field '__provider' on 'd1.QueryState'
```

Without a state migration the alias retype does not diff to nothing; it fails outright at the first preview.

## Confirmed, not assumed

The gate test was written **first** and run **before** the migration existed. It failed with exactly that message against the real gRPC path (`integration.NewServer` → `srv.Diff`) — so the gap is demonstrated, not inferred from the migration plan.

The state shape is read off the live stack rather than imagined:

```
$ pulumi stack export -C deploy/services/uptime -s prod
urn:     …UptimeMonitor$pulumi-nodejs:dynamic:Resource::uptime-prod-schema
outputs: __provider, accountId, apiToken, databaseId, sql, sqlHash
```

## Why this is simpler than litellm's

litellm needed **two** shapes per resource to discriminate a sentinel-encoded optional (`maxBudget: number | ""`) — one shape alone would make state carrying a real budget fail both the migrator *and* the normal decode.

d1 has no such hazard: `accountId`, `databaseId`, `sql` and `apiToken` are all strings, so an empty value is a valid value of its own type and decodes fine either way. `__provider` is the only obstacle, so one shape suffices.

## Tests

| test | what it pins |
|---|---|
| `TestDiffAcceptsLiveDynamicProviderState` | **the gate** — live state in, retyped inputs out, must diff to nothing |
| `TestRealSqlChangeSurvivesTheMigration` | dropping `__provider` is the ONLY permitted effect; an edited schema must still plan an update, not be silently swallowed |
| `TestPluginNativeStateStillDecodes` | state with no `__provider` decodes by the normal path — the migration is additive-safe |

Schema verified unchanged via `pulumi package get-schema`: `__provider` stays migration-only and never enters the public surface.

## This gap is systemic

Only **litellm** had `__provider` handling. `attic`, `onepassword`, `matrix` and `r2` all still need the same treatment before their families can be retyped — every one would hit an identical fatal decode on its first preview.

Doing d1 first because uptime/prod is the smallest live blast radius; the pattern here is what the others follow.

## Not in this PR

The TypeScript retype of `D1Query` and the alias cutover on uptime. This is the provider-side prerequisite only.

## Test plan

- [x] `nix build .#checks.x86_64-linux.provider-go-test` — pass (gofmt + `go test -race ./...`)
- [x] Gate test verified failing **before** the migration, passing after
- [x] `pulumi package get-schema` — d1 surface unchanged

Known-inherited, unchanged: #345 (stale `pnpmDepsHash`), #347 (treefmt). This PR touches no TypeScript.

### DIFF
--- a/provider/d1/migrations.go
+++ b/provider/d1/migrations.go
@@ -37,7 +37,13 @@ type queryStateV0 struct {
 	// Provider is the serialized dynamic-provider closure. Declared only so it
 	// can be dropped — nothing reads it, and this is how __provider finally
 	// leaves state.
-	Provider string `pulumi:"__provider,optional"`
+	//
+	// Deliberately NOT optional, and that tag is load-bearing. infer tries this
+	// migrator before the normal decode, so `optional` would make the shape
+	// match plugin-native state too and route every future read through the
+	// migration — the opposite of the additive-safety described above. Required
+	// is what makes native state miss this shape and fall through.
+	Provider string `pulumi:"__provider"`
 }
 
 // StateMigrations satisfies infer.CustomStateMigrations[QueryState].
--- a/provider/d1/migrations.go
+++ b/provider/d1/migrations.go
@@ -1,0 +1,55 @@
+package d1
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi-go-provider/infer"
+)
+
+// State migration off the dynamic provider.
+//
+// Live state written by the dynamic provider carries `__provider`, the
+// serialized JavaScript closure — the very thing this plugin exists to stop
+// storing. No struct here declares it, and infer treats an unrecognized
+// property as a decode FAILURE rather than something quietly ignored:
+//
+//	Unrecognized field '__provider' on 'd1.QueryState'
+//
+// So without this migration the alias retype does not diff to nothing; it
+// fails outright at the first preview. Verified against the real gRPC path in
+// migrations_test.go using the shape read off uptime/prod.
+//
+// This is simpler than the litellm equivalent, and deliberately so. litellm
+// needed two shapes per resource to discriminate a sentinel-encoded optional
+// (`maxBudget: number | ""`), because a single shape would have made state
+// carrying a real budget fail both the migrator and the normal decode. d1 has
+// no such hazard: accountId, databaseId, sql and apiToken are all strings, so
+// an empty value is a valid value of its own type and decodes fine either way.
+// __provider is the only obstacle, so one shape suffices.
+//
+// The migration is additive-safe: state that never passed through the dynamic
+// provider has no __provider, fails this migrator's decode ("doesn't fit into
+// the migrator", per infer's own docs), and falls through to the normal path
+// unchanged.
+type queryStateV0 struct {
+	QueryArgs
+	SQLHash string `pulumi:"sqlHash,optional"`
+	// Provider is the serialized dynamic-provider closure. Declared only so it
+	// can be dropped — nothing reads it, and this is how __provider finally
+	// leaves state.
+	Provider string `pulumi:"__provider,optional"`
+}
+
+// StateMigrations satisfies infer.CustomStateMigrations[QueryState].
+func (Query) StateMigrations(context.Context) []infer.StateMigrationFunc[QueryState] {
+	return []infer.StateMigrationFunc[QueryState]{
+		infer.StateMigration(func(_ context.Context, old queryStateV0) (infer.MigrationResult[QueryState], error) {
+			return infer.MigrationResult[QueryState]{
+				Result: &QueryState{
+					QueryArgs: old.QueryArgs,
+					SQLHash:   old.SQLHash,
+				},
+			}, nil
+		}),
+	}
+}
--- a/provider/d1/migrations_test.go
+++ b/provider/d1/migrations_test.go
@@ -1,0 +1,132 @@
+package d1
+
+import (
+	"context"
+	"testing"
+
+	"github.com/blang/semver"
+	p "github.com/pulumi/pulumi-go-provider"
+	"github.com/pulumi/pulumi-go-provider/infer"
+	"github.com/pulumi/pulumi-go-provider/integration"
+	"github.com/pulumi/pulumi/sdk/v3/go/property"
+)
+
+// liveQueryState mirrors what uptime/prod actually holds today, read off the
+// stack rather than imagined:
+//
+//	pulumi stack export -C deploy/services/uptime -s prod
+//	  urn: …:UptimeMonitor$pulumi-nodejs:dynamic:Resource::uptime-prod-schema
+//	  outputs: __provider, accountId, apiToken, databaseId, sql, sqlHash
+//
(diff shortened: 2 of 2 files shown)


jmmaloney4:
## Review triage — `Logic Mismatch on Optional Tag` confirmed and fixed

PR-Agent's focus area was **correct**, and it matched a mismatch I had independently derived from the `infer` source. Fixed in 98643cf.

### The finding

`hydrateFromState` ([`infer/resource.go:1550`](https://github.com/pulumi/pulumi-go-provider/blob/v1.4.1/infer/resource.go)) runs migrations **before** the normal decode, and `migrateState` skips a migrator only when `ende.DecodeAny` into its old shape fails. With `__provider` tagged `optional`, `queryStateV0` decoded plugin-native state cleanly too — so the migration fired on **every** state read, not just legacy state. That directly contradicted the file's own comment and the PR body's claim that native state "falls through to the normal path".

Outcome was benign (the migration was an identity copy), but the documented mechanism was false, and the blast radius was permanent rather than legacy-only.

### Discriminating experiment

The three existing tests pass under *both* tags, so they did not pin this. To distinguish the paths I temporarily made the migration non-identity (`SQL = "MANGLED-BY-MIGRATION"`) and ran `nix build .#checks.x86_64-linux.provider-go-test`:

| v0 `__provider` tag | `TestDiffAcceptsLiveDynamicProviderState` | `TestPluginNativeStateStillDecodes` |
|---|---|---|
| `optional` (as opened) | FAIL (migrated) | **FAIL (migrated — fall-through did not happen)** |
| required (98643cf) | FAIL (migrated) | **PASS (fell through)** |

Required is what makes the documented additive-safety real. Added a comment marking the tag as load-bearing so it is not "helpfully" re-optionalised.

No schema impact: `queryStateV0` is a migration-only reflection shape, never registered via `infer.Resource`.

### Negative control (verifying the migration actually fires)

Renamed `StateMigrations` so it no longer satisfies `infer.CustomStateMigrations`, staged, rebuilt:

```
--- FAIL: TestDiffAcceptsLiveDynamicProviderState
        __provider: Unrecognized field '__provider' on 'd1.QueryState'
--- FAIL: TestRealSqlChangeSurvivesTheMigration
        __provider: Unrecognized field '__provider' on 'd1.QueryState'
```

The gate fails with exactly the claimed message. The migration is genuinely wired: `provider.go:97` registers `infer.Resource(d1.Query{})` as a **value**, matching the value receiver on `StateMigrations` and the test server's `infer.Resource(Query{})` — so there is no silent never-runs path.

### One-shape claim — holds

`QueryArgs` is `AccountID`, `DatabaseID`, `SQL`, `APIToken`, all plain non-optional `string`; `QueryState` adds `SQLHash string`. No pointer/typed optional anywhere, so there is no sentinel-encoded field that could fail both the migrator and the normal decode the way litellm's `maxBudget *float64` does. One shape is correct.

`sqlHash,optional` in the v0 shape is also right: live state carries it, but optional is strictly the safer choice — a required tag would make state lacking it miss the migrator and then hard-fail the normal decode on `__provider`.

### Secret handling — checked, no issue

`apiToken` is `provider:"secret"`, and `migrateState` carries a caveat about dropping secrets. Not reachable here: `Diff` and `Update` discard the migration encoder and re-derive it from `ende.Decode[I](req.Inputs)`, and `d1.Query` implements no `CustomRead`, so the one call site that *does* use the state encoder returns `req.Properties` unmodified.

jmmaloney4:
## Merge gate audit

**HEAD:** `98643cf4c0ff242d65b1dedd9b572dbfe89bb29d` · `mergeable=MERGEABLE` · `mergeStateStatus=UNSTABLE`

### Review convergence — converged at current HEAD

| gate | state |
|---|---|
| `/review` persistent | `(Review updated until commit …/98643cf)` — **"No major issues detected"**. The `Logic Mismatch on Optional Tag` focus area raised on `4e7df5d` is fixed and gone. |
| `/improve` | Manually re-triggered after the push; resolved at 18:58:50 on `98643cf` → **"No code suggestions found for the PR."** |

One actionable finding was raised across the whole loop, and it was **real** — not merged over. Triage, discriminating experiment, and fix are documented in [the triage comment](https://github.com/jmmaloney4/sector7/pull/353#issuecomment-5170512895); fixed in `98643cf`.

### Local validation (the gate — see below)

```
nix build .#checks.x86_64-linux.provider-go-test --no-link   # gofmt + go test -race ./...
EXIT=0
```

Run on the committed tree at `98643cf` with a clean working tree (no `uncommitted changes` warning), so Nix tested the real PR content rather than a stale snapshot.

Additionally verified, beyond the PR's own claims:
- **Negative control** — with `StateMigrations` renamed so it no longer satisfies `infer.CustomStateMigrations`, the gate test fails with exactly `Unrecognized field '__provider' on 'd1.QueryState'`. The migration is genuinely wired, not passing for the wrong reason.
- **Path discrimination** — a temporarily non-identity migration proves legacy state migrates and plugin-native state now falls through, which the three shipped tests could not distinguish on their own.
- **One-shape claim** — `QueryArgs`/`QueryState` are all plain non-optional `string`; no typed optional exists that could reproduce litellm's `maxBudget *float64` sentinel hazard.

### CI — not a gate here

All three self-hosted runners are `offline`:

```
jmmaloney4-sector7-6kwsg-runner-45mjt   offline
jmmaloney4-sector7-6kwsg-runner-blvb8   offline
jmmaloney4-sector7-6kwsg-runner-ff7nm   offline
```

The single check (`adr-management / push placeholder to main`) is `QUEUED` and will not start. This is the SOP's `UNSTABLE` + self-hosted-queue case: `UNSTABLE` reflects the never-scheduled job, not a failure, so local validation is the merge gate.

### Inherited failures — confirmed, not introduced, not fixed here

- **#345** (stale `pnpmDepsHash` → `tsc`/`vitest`) — OPEN. This PR touches no TypeScript.
- **#347** (treefmt red on main, 7 files) — OPEN. Reproduced on this branch; the failing set is exactly:
  `docs/renovate/README.md`, `docs/renovate/workflows.md`, `packages/sector7/pulumi-config/index.ts`, `packages/sector7/pulumi-config/op-secret-helpers.ts`, `packages/sector7/tests/op-secret-helpers.test.ts`, `packages/sector7/tests/port-forward.test.ts`, `renovate/minor-patch-automerge.json`.
  Zero overlap with this diff (`provider/d1/migrations.go`, `provider/d1/migrations_test.go`); `provider/d1` appears nowhere in the treefmt log. Go formatting for this PR is covered by the gofmt stage of `provider-go-test`, which is green.

### Remaining

No unresolved correctness, security, data-loss, or architectural concerns. Merging with a merge commit (squash and rebase are disabled on this repo).

PR #354, opened 2026-08-03
test(d1): guard the __provider tag that decides when the migration runs

jmmaloney4:
## Summary

Follow-up to #353. `98643cf` fixed a real bug in that PR during review — but nothing guards the fix.

`__provider` had been tagged `,optional` in `queryStateV0`. Because `infer` runs migrations **before** the normal decode and skips a migrator only when decoding into its old shape *fails*, `,optional` made the shape match **plugin-native state too** — state that never carried the property at all. So the migration fired on **every** state read, permanently, rather than only on state written by the dynamic provider. That is the exact opposite of what the migration's own comment promised.

Benign in outcome, because the migration is an identity copy. But the documented mechanism was false and the scope was permanent rather than legacy-only.

## Why this is a structural assertion, not a behavioural one

**All three tests from #353 pass under both tags.** I verified that by reintroducing `,optional` on a throwaway worktree and re-running the suite — everything green. An identity migration is unobservable from outside, so no amount of `Diff`-level testing can distinguish "migration ran" from "migration was skipped."

Distinguishing them requires a deliberately non-identity migration, which isn't something to ship in production code just to make the property testable.

So this asserts the invariant directly: the tag must be exactly `pulumi:"__provider"`. The failure message explains the consequence rather than just reporting a mismatch:

```
queryStateV0.Provider must be tagged exactly `pulumi:"__provider"`, got "__provider,optional".
Adding ,optional makes this shape match plugin-native state as well, so the
migration would run on every state read instead of only on state written by
the dynamic provider.
```

The realistic regression is someone tidying the struct and adding `,optional` for consistency with the neighbouring optional fields — which reads as harmless and isn't.

## Verified both directions

- passes on the fixed tree
- reintroducing `,optional` fails it with the message above

## Relevance beyond d1

The same shape is about to be written four more times — `attic`, `onepassword`, `matrix`, `r2` all still need `__provider` migrations before their families can be retyped. This guard is the pattern each should carry, and the mistake is clearly an easy one to make: I made it on the first attempt.

## Test plan

- [x] Negative control — fails correctly

Known-inherited, untouched: #345, #347. No TypeScript in this diff.

### DIFF
--- a/provider/d1/migrations_test.go
+++ b/provider/d1/migrations_test.go
@@ -2,6 +2,7 @@ package d1
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	"github.com/blang/semver"
@@ -128,5 +129,44 @@ func TestPluginNativeStateStillDecodes(t *testing.T) {
 	}
 	if resp.HasChanges {
 		t.Fatalf("unchanged plugin-native state must diff to nothing; got %+v", resp.DetailedDiff)
+	}
+}
+
+// The `__provider` tag must stay REQUIRED, and this is a structural assertion
+// rather than a behavioural one for a specific reason.
+//
+// infer runs migrations BEFORE the normal decode, and skips a migrator only
+// when decoding into its old shape FAILS. Tagging `__provider` as `optional`
+// therefore makes queryStateV0 match plugin-native state too — state that
+// never carried the property at all — so the migration fires on EVERY read,
+// permanently, instead of only on legacy state. That is the opposite of what
+// the migration's own comment promises.
+//
+// The original version of this file shipped with `,optional` and was corrected
+// in 98643cf. The outcome was benign only because the migration is an identity
+// copy; the documented mechanism was wrong and the scope was permanent rather
+// than legacy-only.
+//
+// A behavioural test cannot catch this. All three tests above pass under BOTH
+// tags — verified by reintroducing `,optional` and re-running them — precisely
+// because an identity migration is unobservable. Distinguishing the two paths
+// requires a deliberately non-identity migration, which is not something to
+// ship in production code just to make it testable.
+//
+// So this asserts the invariant directly: the tag is load-bearing, and a
+// future "cleanup" that adds `,optional` for consistency with the other
+// optional fields silently changes when the migration runs.
+func TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState(t *testing.T) {
+	f, ok := reflect.TypeOf(queryStateV0{}).FieldByName("Provider")
+	if !ok {
+		t.Fatal("queryStateV0.Provider is gone; if the field was renamed, update this guard rather than deleting it")
+	}
+
+	tag := f.Tag.Get("pulumi")
+	if tag != "__provider" {
+		t.Fatalf("queryStateV0.Provider must be tagged exactly `pulumi:\"__provider\"`, got %q.\n"+
+			"Adding ,optional makes this shape match plugin-native state as well, so the\n"+
+			"migration would run on every state read instead of only on state written by\n"+
+			"the dynamic provider.", tag)
 	}
 }
--- a/provider/d1/migrations_test.go
+++ b/provider/d1/migrations_test.go
@@ -143,18 +143,33 @@ func TestPluginNativeStateStillDecodes(t *testing.T) {
 // the migration's own comment promises.
 //
 // The original version of this file shipped with `,optional` and was corrected
-// in 98643cf. The outcome was benign only because the migration is an identity
-// copy; the documented mechanism was wrong and the scope was permanent rather
-// than legacy-only.
+// in 98643cf. On the states exercised above the outcome was benign, because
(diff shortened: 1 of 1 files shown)


jmmaloney4:
## Review finding: the negative control holds, but the stated rationale did not

**Negative control — verified.** Reintroduced `,optional` on `queryStateV0.Provider` in `provider/d1/migrations.go` and ran `nix build .#checks.x86_64-linux.provider-go-test`:

```
--- FAIL: TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState (0.00s)
    migrations_test.go:167: queryStateV0.Provider must be tagged exactly `pulumi:"__provider"`, got "__provider,optional".
```

The guard fails, with the explanatory message, for the right reason. The other three tests passed in that same run — confirming the PR's claim that they do not discriminate. Restored afterwards; `provider-go-test` is green on the current HEAD.

**But "a behavioural test cannot catch this" was not true.** The comment justified the structural assertion by asserting that an identity migration is unobservable, so no behavioural test could exist. I probed that rather than taking it, and found a discriminating state.

`sqlHash` is `,optional` on `queryStateV0` but **required** on `QueryState`. The v0 shape is therefore strictly more permissive than the type it migrates to, and plugin-native state missing `sqlHash` diverges:

| state | `pulumi:"__provider"` | `pulumi:"__provider,optional"` |
|---|---|---|
| native, no `sqlHash` | **`Diff` errors** (fails both decodes) | **succeeds**, `SQLHash: ""`, diffs to nothing |
| legacy (`__provider`), no `sqlHash` | succeeds, no changes | succeeds, no changes |
| native + unknown extra property | errors | errors |

Measured, not reasoned — throwaway probe tests under both tags via the same flake check.

Two consequences:

1. **The bug was less benign than the comment said.** `,optional` did not only mis-scope the migration to fire on every read; it also converted a decode error into a silent empty hash for state missing `sqlHash`. Worth knowing, since `sqlHash` is what change detection keys on.
2. **The structural assertion is still the right call** — but for a better reason than "no behavioural test exists". The divergence above is an accident of the current field set, not designed behaviour: it discriminates only while some field is optional on `queryStateV0` and required on `QueryState`, it silently stops discriminating the moment that ceases to hold, and a behavioural test would pin an error path nobody chose (`Diff` must reject state with no `sqlHash`) as though it were the point. Its failure message — `expected an error, got nil` — would also tell the next reader nothing about `__provider`. A guard that quietly stops guarding is the exact failure mode this PR exists to prevent.

`a4f1366` rewrites the rationale accordingly. The assertion itself is unchanged; only the comment moved, from an overclaim that a future reader could reasonably act on ("the tag is behaviourally inert, drop the guard") to the measured reason structural is preferred.

jmmaloney4:
## Merge-gate audit — `a4f1366`

**Review convergence (Phase 5).**
- `/review` persistent comment explicitly updated to `a4f136604a31640421a997a37c4cddd0aa733160` = current HEAD. Clean: *No security concerns identified*, *No major issues detected*, no non-empty findings sections.
- `/improve` clean on this HEAD. The push-triggered run posted `No code suggestions found for the PR.`; because clean runs carry no `<!-- sha -->` marker, I did not rely on recency — I fired a manual `/improve` and it returned `No code suggestions found for the PR.` against the current HEAD.
- One finding raised and resolved during this pass, by me rather than the bot: the guard's justification comment overclaimed. Detail and measurements in the [preceding comment](https://github.com/jmmaloney4/sector7/pull/354#issuecomment-5170851584); `a4f1366` corrects the rationale. The assertion itself is unchanged from `5f5cf43`.

**Negative control — the point of this PR.**
Reintroduced `,optional` on `queryStateV0.Provider`, rebuilt, and confirmed the guard fails for the right reason with its explanatory message:

```
--- FAIL: TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState (0.00s)
    migrations_test.go:167: queryStateV0.Provider must be tagged exactly `pulumi:"__provider"`, got "__provider,optional".
```

The three behavioural tests from #353 passed in that same run, confirming they do not discriminate. `provider/d1/migrations.go` restored; working tree clean at HEAD.

**Validation (Phase 6).** `nix build .#checks.x86_64-linux.provider-go-test --no-link` → exit 0 on `a4f1366`, run unpiped, with the worktree clean and HEAD equal to the PR head. All provider packages `ok`, including `provider/d1`.

**CI.** `adr-management` is `QUEUED`; all three self-hosted runners report `status: offline`, so checks cannot start. `mergeStateStatus=UNSTABLE` is entirely attributable to that queue — `mergeable=MERGEABLE`. Per the offline-queue rule, local validation is the gate.

**Scope.** Diff is one Go test file, `provider/d1/migrations_test.go`. No TypeScript, so the known-inherited #345 (stale `pnpmDepsHash`) and #347 (treefmt drift) are untouched and not blockers here. `nix fmt` run on the changed file; no spillover.

Merging with a merge commit (`squashMergeAllowed=false`, `rebaseMergeAllowed=false` on this repo).

PR #355, opened 2026-08-03
feat(d1): retype D1Query onto the sector7 resource plugin

jmmaloney4:
## Summary

The TypeScript half of the uptime slice. Replaces `D1Query`'s Pulumi dynamic provider with the plugin-backed token `sector7:d1:Query`, plus the alias that makes the cutover a URN rewrite rather than a create/delete.

The CRUD implementation already existed in `provider/d1/query.go` — this wires the wrapper to it.

`d1-query.ts` drops **327 → 95 lines**. The check/diff/create/update/delete callbacks, the Cloudflare REST client, and the dynamic-`import()` dance for `node:crypto` all move to Go. That last one is worth noting: it existed *only* because Pulumi serialised this closure into state and top-level Node built-in imports capture unserialisable native functions. A plugin has its own process, so the constraint evaporates.

`DynamicResourceOptions` and `rejectCloudProviderOptions` go too — their premise inverts under the plugin, where `provider` is a meaningful option rather than something that routes a resource through the wrong bridge. Same treatment litellm's retype applied. Nothing outside this module consumed the exported type.

## Fidelity check

I compared the Go `Diff` against the TypeScript it replaces, including the detail most likely to drift silently:

| property | TypeScript | Go |
|---|---|---|
| `accountId`, `databaseId`, `sql` | replace | `UpdateReplace` |
| `apiToken` | update (in place) | `Update` |
| `deleteBeforeReplace` | `replaces.length > 0` | `replaces > 0` |

Identical, including that a token rotation must **not** replace.

## No coverage lost

The Go side had **no `Check` test at all**, so deleting the TypeScript one would have dropped that behaviour silently. Added:

- `TestCheck` — carries the original `it(...)` strings verbatim (`reports no check failures for valid inputs`, `reports check failures when required fields are missing`)
- `TestDiffIsANoOpWhenNothingChanged` — the case that matters on every routine `up`
- a **whitespace-only SQL** case. I checked that Go's `Check` uses `strings.TrimSpace`, matching the TypeScript's `sql.trim().length === 0` — blank SQL is a no-op that would still record a `sqlHash`, so the resource would claim a schema was applied when nothing ran.

## The TypeScript test is rewritten, not deleted

Registration is still TypeScript and still worth pinning. Five cases: the resource token, input forwarding, the dynamic-provider alias, the plugin version, and that a caller-supplied `provider` is now **passed through rather than rejected** (the inverted guard).

Its header maps every moved `it(...)` string to the Go subtest that inherited it, so the two lists can be diffed mechanically during review.

## Depends on #353

Without that state migration this retype fails at first preview with `Unrecognized field '__provider' on 'd1.QueryState'` rather than diffing to nothing. #353 is merged; this is safe to land after it.

## Test plan

- [x] `tsc -p tsconfig.build.json --noEmit` — clean
- [x] `vitest run tests/d1-query.test.ts` — 5/5 pass
- [x] Full `vitest run` — **301 pass**; the delta from 304 is exactly the 8 moved d1 tests replaced by 5
- [x] `treefmt` — flags only `renovate/minor-patch-automerge.json` (#347), none of my files

Sole full-suite failure is the known-inherited `tests/renovate-pulumi.test.ts`, previously reproduced on a clean `origin/main` worktree.

## Not in this PR

The alias cutover on `uptime` itself — a garden-side version bump plus `pulumi preview`/`up` against the live stacks. That's the next step, and `uptime/prod` holds exactly one live instance.


jmmaloney4:
## Merge-gate audit

**Head:** `258b4e8573175adb7448b41ca98ad08cfa1d9f84`

### Review convergence (Phase 5)
- `/review` — marker `(Review updated until commit .../258b4e8...)` equals current HEAD. *No security concerns identified*, *No major issues detected*.
- `/improve` — two fresh runs after the review-fix push (20:23:02 push-triggered, 20:23:42 manual), both `No code suggestions found for the PR.`

### What review changed
One real coverage gap found and fixed in `258b4e8`. The deleted TypeScript case `detects changes when apiToken is rotated` asserted three things — `changes`, an **empty** `replaces` list, and `deleteBeforeReplace === false`. The Go port kept only the first and a narrower form of the second (it checked `apiToken`'s own diff kind, nothing else). A regression that set `DeleteBeforeReplace` unconditionally, or that marked an unrelated property for replacement, would have passed the Go suite. Both assertions are now restored.

The same commit splits `TestDiffReplacesOnStatementIdentityButNotOnToken` into subtests carrying the exact `it(...)` strings. The file header claimed every moved string was "carried verbatim as a Go subtest name"; three of the six appeared nowhere in the Go file, so the promised mechanical diff did not actually work. All six now grep clean, and the header is corrected to note that one case lands in a doc comment rather than a subtest name.

### Behavioural equivalence, Go `Diff` vs the deleted TypeScript `diff`
Verified against `HEAD~1:packages/sector7/d1/d1-query.ts`, not taken on trust. Equivalent on every axis:

| | TypeScript | Go |
|---|---|---|
| replace set | `replaces` gets `accountId`, `databaseId`, `sql` | same three at `Kind: UpdateReplace` |
| token change | `hasTokenChange`, not in `replaces` | `apiToken` at `Kind: Update` |
| changed? | `replaces.length > 0 \|\| hasTokenChange` | `len(diffs) > 0` — same predicate, the map only gains entries for those four |
| deleteBeforeReplace | `replaces.length > 0` | `replaces > 0`, counting only `UpdateReplace` kinds |

`sqlHash` is diffed by neither (absent from `QueryArgs`, as it was from the TS comparison). `Check` is equivalent too: `TrimSpace(SQL) == ""` covers the old `!sql || sql.trim().length === 0`, and the three required-field checks are unchanged. The restored `DeleteBeforeReplace == false` assertion passing is direct empirical confirmation of the token-rotation row.

### Coverage audit of the 8 deleted TypeScript cases
Six moved and genuinely assert the same behaviour (two are *stronger*: the missing-fields case now pins all four properties rather than `length >= 1`, and whitespace-only SQL is newly covered). One — `registers a dynamic resource with correct args` — is replaced by two TS registration tests. The eighth, `rejects cloud provider options`, is deliberately inverted rather than dropped: `provider`/`providers` are meaningful under the plugin, and a new test pins that they pass through. Nothing was silently lost.

### Alias and version wiring
Token `sector7:d1:Query`, alias `[{ type: "pulumi-nodejs:dynamic:Resource" }]`, `version: PLUGIN_VERSION` — all three confirmed byte-identical to `pluginOpts()` in `packages/sector7/litellm/admin.ts`, the wiring already shipped and exercised against litellm/prod. The token is corroborated independently by the URNs in `provider/d1/migrations_test.go`, which drive the real gRPC path. `{ sqlHash: undefined, ...args }` matches the shipped `{ teamId: undefined, ... }` / `{ tokenId: undefined, ... }` pattern.

Removing the exported `DynamicResourceOptions` from `packages/sector7/d1/` breaks nothing: the type is declared per-module (`onepassword`, `r2`, `attic` each keep their own), and no consumer in this repo or in garden imports it from `/d1`.

### Validation (Phase 6)
| Check | Result |
|---|---|
| `checks.x86_64-linux.provider-go-test` | **pass** — all 8 packages `ok`, incl. `provider/d1` |
| `tsc -p tsconfig.build.json --noEmit` | **pass** |
| `vitest run` | **301 passed**, exactly as predicted (304 − 8 removed CRUD + 5 new registration) |
| `checks.x86_64-linux.treefmt` | PR-scope **clean** |

The `tsc`/`vitest` Nix checks remain blocked by the stale `pnpmDepsHash` (#345), unrelated to this PR, so TypeScript was validated directly via `nix develop .#ci-pnpm`.

**Two treefmt regressions this branch introduced were found and fixed** — biome import ordering in `d1-query.ts` and markdown table padding in `README.md`. After the fix, the failing set on this HEAD is *exactly* the inherited set on a fresh `origin/main` worktree at `0c7b012`, proven by running the same derivation on both:

```
docs/renovate/README.md
docs/renovate/workflows.md
packages/sector7/pulumi-config/index.ts
packages/sector7/pulumi-config/op-secret-helpers.ts
packages/sector7/tests/op-secret-helpers.test.ts
packages/sector7/tests/port-forward.test.ts
renovate/minor-patch-automerge.json
```

Seven files, identical on both sides — repo-wide debt (broader than the #347 note, which named only the renovate preset), to be cleared in a separate mechanical PR. `tests/renovate-pulumi.test.ts` likewise fails identically on clean `origin/main`.

### Remaining gate items
`mergeable=MERGEABLE`. `mergeStateStatus=UNSTABLE` solely because `add-to-project` and `adr-management` are `QUEUED` on self-hosted runners that are offline — local validation is the gate per the offline-queue rule. No rebase was needed: main moved only in `provider/d1/migrations_test.go` (#353/#354), disjoint from this diff.

Merging with a merge commit (the only method this repo allows).

PR #356, opened 2026-08-04
fix(onepassword): keep secret field values out of state, and migrate __provider

jmmaloney4:
## Summary

Two defects with one root cause: `ItemState` embedded `ItemArgs`.

### 1. Security — secret field values were being persisted

The dynamic provider this replaces **never** wrote field values into state. Its `stateOuts` listed the connection and identity inputs explicitly and stopped, keeping only `managedLabels` (labels, not values) and `contentHash` (a digest over the values):

```ts
return { kubeconfig, connectToken, namespace, deploymentName, connectPort,
         vault, title, category, uuid, itemPath, contentHash,
         managedLabels: i.fields.map((f) => f.label) };   // no `fields`
```

Embedding `ItemArgs` silently reversed that. `Create` and `Update` both write `ItemState{ItemArgs: a, …}`, so **every managed secret would be persisted**. Encrypted — `Field.Value` is tagged secret — but present, which is a weaker property than absent. The reason `contentHash` is *itself* marked secret is precisely to avoid handing out an offline oracle over those values; storing the values alongside it defeats the intent.

### 2. Migration — live state cannot decode

The same embedding made `ItemState` require `fields`, which live state does not have. The five dynamic `OnePasswordItem`s in gateway/ergon/matrix/litellm carry no such key, so the alias retype could not have decoded them *even before* reaching `__provider`.

## Fix

`ItemState` now lists its properties explicitly, mirroring `stateOuts`. Verified with `pulumi package get-schema` that the output set is exactly the 12 keys live state holds, and that `fields` remains an **input**:

| | |
|---|---|
| outputs | `category, connectPort, connectToken, contentHash, deploymentName, itemPath, kubeconfig, managedLabels, namespace, title, uuid, vault` |
| inputs | …same, minus the four computed, **plus `fields`** |

That is byte-identical to live state except for `__provider`, which the new migration drops.

Nothing read persisted `Fields` — `Diff` hashes the *incoming* fields against the stored `contentHash`, and `Update` reconciles removals from `ManagedLabels`, both matching the TypeScript. `ItemState.target()` rebuilds the connection half for `Delete`.

Zero live instances are typed `sector7:onepassword:Item` yet, so this schema change costs nothing.

## Tests — each verified to fail when its subject is broken

| test | negative control |
|---|---|
| `TestDiffAcceptsLiveDynamicProviderState` | remove `StateMigrations` → `Unrecognized field '__provider'` ✓ |
| `TestRealFieldChangeSurvivesTheMigration` | a rotated secret must still plan a change |
| `TestPluginNativeStateStillDecodes` | additive-safety |
| `TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState` | the load-bearing tag, as in d1 |
| `TestItemStateNeverCarriesFieldValues` | add an explicit `Fields` → fires with a legible message ✓ |

The fixture is gateway/prod's `mcp-consumer-tokens`, read off the stack rather than invented, and it **derives** `contentHash` from the same fields the inputs carry so it cannot drift into disagreeing with itself. My first draft hardcoded a placeholder hash and failed for exactly that reason — a useful failure, and now structurally impossible.

## One thing the comment is careful about

The two ways `TestItemStateNeverCarriesFieldValues` can regress fail *differently*, and the comment says so rather than overclaiming:

- **explicit `Fields` field** → caught by the assertion, with an explanatory message
- **re-embedding `ItemArgs`** → caught earlier and less legibly, by `infer` panicking with `could not annotate field: could not find field`, which aborts the test binary before the assertion runs

Protected either way; only one explains itself. I verified both paths rather than assuming the assertion covered them.

## Test plan

- [x] `nix build .#checks.x86_64-linux.provider-go-test` — pass (all 8 packages)
- [x] Both negative controls confirmed
- [x] `pulumi package get-schema` — output/input sets match live state

## Not in this PR

The TypeScript retype of `OnePasswordItem` and the alias cutover across gateway → ergon → matrix → litellm. This is the provider-side prerequisite.

### DIFF
--- a/provider/onepassword/item.go
+++ b/provider/onepassword/item.go
@@ -106,9 +106,15 @@ type ItemState struct {
 }
 
 // target rebuilds the connection half of the inputs from stored state, for the
-// paths that must reach Connect using what was persisted rather than what the
-// program currently declares (Delete, and Update's removal reconciliation).
-// Fields is intentionally absent — nothing on those paths needs it.
+// one path that must reach Connect using what was persisted rather than what
+// the program currently declares: Delete, which is handed no inputs at all.
+// Fields is intentionally absent — nothing on that path needs it.
+//
+// Update is deliberately NOT a caller. It connects with the CURRENT inputs, so
+// that a token rotation or cluster move takes effect on the same apply that
+// declares it; the only thing it draws from prior state is ManagedLabels, for
+// removal reconciliation. Delete is the sole caller — keep it that way, or the
+// "what was persisted" framing above stops being true.
 func (s ItemState) target() ItemArgs {
 	return ItemArgs{
 		Kubeconfig:     s.Kubeconfig,
--- a/provider/onepassword/item.go
+++ b/provider/onepassword/item.go
@@ -61,8 +61,40 @@ type ItemArgs struct {
 	Fields         []Field `pulumi:"fields"`
 }
 
+// ItemState deliberately does NOT embed ItemArgs, and so deliberately does not
+// carry Fields.
+//
+// The dynamic provider this replaces never persisted field values. Its
+// stateOuts (packages/sector7/onepassword/item.ts) listed the connection and
+// identity inputs explicitly and stopped there, keeping only `managedLabels`
+// (labels, not values) and `contentHash` (a digest over the values) — secret
+// values never entered state at all. Embedding ItemArgs here silently reversed
+// that: Create and Update both write ItemState{ItemArgs: a, …}, so every
+// managed secret would be persisted. Encrypted, since Field.Value is tagged
+// secret — but present, which is a weaker property than absent, and the reason
+// contentHash is itself marked secret is precisely to avoid handing out an
+// offline oracle over those values.
+//
+// It also matches what live state actually looks like. The five dynamic
+// OnePasswordItems in gateway/ergon/matrix/litellm have no `fields` key in
+// their outputs, so an ItemState requiring one cannot decode them and the alias
+// retype would fail before it began.
+//
+// Nothing reads persisted Fields: Diff hashes the INCOMING fields and compares
+// against the stored contentHash, and Update reconciles removals from
+// ManagedLabels. Both match the TypeScript.
 type ItemState struct {
-	ItemArgs
+	// Connection and identity inputs are echoed back, mirroring stateOuts in
+	// the TypeScript. Fields is the deliberate omission.
+	Kubeconfig     string `pulumi:"kubeconfig,optional" provider:"secret"`
+	ConnectToken   string `pulumi:"connectToken" provider:"secret"`
+	Namespace      string `pulumi:"namespace"`
+	DeploymentName string `pulumi:"deploymentName,optional"`
+	ConnectPort    int    `pulumi:"connectPort,optional"`
+	Vault          string `pulumi:"vault"`
+	Title          string `pulumi:"title"`
+	Category       string `pulumi:"category,optional"`
+
 	UUID     string `pulumi:"uuid"`
 	ItemPath string `pulumi:"itemPath"`
 	// ContentHash is secret: it is a SHA-256 over the field values, so exposing
@@ -71,6 +103,23 @@ type ItemState struct {
 	// ManagedLabels records which labels this resource owns, so a later update
 	// can remove ones dropped from input without touching unmanaged fields.
 	ManagedLabels []string `pulumi:"managedLabels"`
+}
+
+// target rebuilds the connection half of the inputs from stored state, for the
+// paths that must reach Connect using what was persisted rather than what the
+// program currently declares (Delete, and Update's removal reconciliation).
+// Fields is intentionally absent — nothing on those paths needs it.
+func (s ItemState) target() ItemArgs {
+	return ItemArgs{
+		Kubeconfig:     s.Kubeconfig,
+		ConnectToken:   s.ConnectToken,
+		Namespace:      s.Namespace,
+		DeploymentName: s.DeploymentName,
+		ConnectPort:    s.ConnectPort,
+		Vault:          s.Vault,
+		Title:          s.Title,
+		Category:       s.Category,
+	}
 }
 
 func (a *ItemArgs) Annotate(ann infer.Annotator) {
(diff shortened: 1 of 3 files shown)


jmmaloney4:
## Merge-gate audit — `33c97b6`

Independent verification of the three load-bearing claims in this PR, plus the standard gate items.

### (a) Schema claim: new `ItemState` output set == live state minus `__provider`

Verified end-to-end rather than by inspection. Built the plugin from this branch
(`nix build .#packages.x86_64-linux.pulumi-resource-sector7`) and read its real schema with
`pulumi package get-schema`, then diffed against the actual live resource
(`pulumi stack export -C deploy/platform/gateway -s prod`,
`urn:pulumi:prod::gateway::pulumi-nodejs:dynamic:Resource::mcp-consumer-tokens`).

```
schema outputs (12): category connectPort connectToken contentHash deploymentName
                     itemPath kubeconfig managedLabels namespace title uuid vault
live outputs   (13): __provider + the same 12
```

Set difference in **both** directions is empty:

| check | result |
|---|---|
| live outputs − `__provider` ⊇ schema outputs | extra in live: `[]` |
| schema outputs ⊇ live outputs − `__provider` | missing from live: `[]` |
| exact match | **True** |

`fields` is confirmed **absent** from live outputs — the PR's central premise holds against the
real stack, not just the fixture. `fields` remains an *input* in the schema
(`requiredInputs: connectToken, fields, namespace, title, vault`), which is what the retype needs.

Secret outputs in the schema: `connectToken`, `contentHash`, `kubeconfig` — no field values anywhere.

One incidental observation, not a defect: live *inputs* lack `connectPort` and `deploymentName`
(the TypeScript never wrote defaults into inputs). Harmless — both are `,optional` with
`ann.SetDefault`, and it is `outputs` that the migration decodes. Both **are** present in live
outputs, which matters for (c) below.

### (b) Nothing read the removed `Fields`

Audited every method on `Item`. The complete set is `Check`, `Diff`, `Create`, `Update`, `Delete`
(plus unexported `connect` / `findItemIDByTitle`) — **there is no `Read` method at all**, so
`pulumi refresh` cannot depend on persisted state either.

| path | state usage | reads persisted `Fields`? |
|---|---|---|
| `Check` | inputs only | no |
| `Diff` | `olds.{Vault,Title,ContentHash,ConnectToken,Kubeconfig,Namespace,DeploymentName,ConnectPort}`; hashes `news.Fields` | no — hashes the **incoming** fields |
| `Create` | `req.Inputs` only | no |
| `Update` | `req.State.ManagedLabels` only | no — reconciles removals from labels |
| `Delete` | `req.State.target()`, `req.State.Vault`, `req.ID` | no |

`grep -n 'State\.Fields\|olds\.Fields\|\.State\.ItemArgs'` over the package returns zero hits
outside a comment. Every one of the 14 `.Fields` references is against `ItemArgs` (incoming), never
`ItemState`. The claim holds.

Additionally checked the reverse risk — that removing `Fields` might orphan *existing* plugin-native
state written by v0.20.x (which would now fail to decode with `Unrecognized field 'fields'`). It
cannot: `sector7:onepassword:Item` is not yet referenced by any TypeScript wrapper, and
`gateway/prod` contains zero resources of that type. The only live instance is still the dynamic one.

### (c) `ItemState.target()` reconstructs everything `connect()` needs

`connect()` consumes exactly five values: `Kubeconfig`, `Namespace`, `DeploymentName`, `ConnectPort`
(→ `kube.Target`) and `ConnectToken` (→ bearer). `target()` carries all five, plus `Vault`/`Title`/
`Category`. `Delete` additionally uses `req.State.Vault` and `req.ID` directly.

The non-obvious risk here was `deploymentName` / `connectPort`: if the dynamic provider had not
echoed its defaults into **outputs**, `target()` would yield `Deployment: ""` / `Port: 0` and
`pulumi destroy` would target no cluster. Confirmed against live state that it did:
`deploymentName: "onepassword-connect"`, `connectPort: 8080`. No missing kubeconfig or namespace either.

### Fixture self-consistency

`liveItemState` derives `contentHash` via `ContentHash("API_CREDENTIAL", managedFields())` from the
same `managedFields()` that `newInputs` puts on `fields`, so the two cannot drift by construction.

Confirmed the guard is *live* rather than vacuous by empirically breaking it — repointed the
fixture hash at a different field set, rebuilt, and got exactly the intended failure:

```
--- FAIL: TestDiffAcceptsLiveDynamicProviderState
    migrations_test.go:126: retype must diff to nothing; got map[fields:{Kind:update InputDiff:false}]
--- FAIL: TestPluginNativeStateStillDecodes
    migrations_test.go:169: unchanged plugin-native state must diff to nothing; got map[fields:...]
FAIL	github.com/jmmaloney4/sector7/provider/onepassword
```

Probe reverted; tree restored to `33c97b6`.

### Review convergence

- `/review` persistent comment updated to `33c97b6` — "No security concerns identified",
  "No major issues detected", no findings sections.
- `/improve` on `33c97b6`: **No code suggestions found for the PR.**
- Gemini is out of daily quota (non-blocking by policy).

One change made during review, comment-only: `target()`'s doc comment claimed it served
"Delete, and Update's removal reconciliation". `target()` has exactly one caller — `Delete`.
`Update` connects with the *current* inputs (so a token rotation applies on the same `up` that
declares it) and draws only `ManagedLabels` from prior state. Corrected in `33c97b6`; no behaviour change.

### Validation

- `nix build .#checks.x86_64-linux.provider-go-test --no-link` → **exit 0** on `33c97b6`
  (fresh derivation, not a cache hit from the prior SHA). `ok github.com/jmmaloney4/sector7/provider/onepassword`,
  all sibling packages green.
- `treefmt` is red, inherited (#347) and **not** PR-scoped: zero `.go` files appear in its output;
  every flagged file is TypeScript. This PR touches no TypeScript.
- `tsc` / `vitest` remain blocked by the stale `pnpmDepsHash` (#345) — likewise not PR-scoped.
- GitHub checks (`add-to-project`, `adr-management`) are `QUEUED` on self-hosted runners that are
  offline; `mergeStateStatus=UNSTABLE` for that reason alone, with `mergeable=MERGEABLE`.
  Per the offline-queue rule, local validation is the gate.

### Non-blocking observation

`Create` under `DryRun` now returns `stateOuts(req.Inputs, "", "")`, so preview shows
`itemPath: "vaults/<vault>/items/"` (trailing empty uuid) where it previously showed `""`.
Both are placeholder values that the real apply overwrites; cosmetic, preview-only, not a regression
in kind. Recording it rather than churning the diff.

**Gate: green.** Merging with a merge commit (`mergeCommitAllowed: true`; squash/rebase disabled).

PR #357, opened 2026-08-04
fix(ci): unbreak `nix flake check` (four independent breakages)

### LINKED ISSUE #345: build: stale pnpmDepsHash breaks all local TypeScript validation

### LINKED ISSUE #347: build: treefmt check is red on main (7 unformatted files)

jmmaloney4:
## Why

`nix flake check` is fully red on `main`, and has been for two weeks. A stale `pnpmDepsHash` made `node_modules` unbuildable, which took **tsc, vitest and pre-commit** with it — so those three have not actually run on any PR since 2026-07-21. Two genuine regressions landed behind that wall.

I hit this while verifying an unrelated change and could not get a green gate to verify against, so this PR is split out on its own.

## What

| Breakage | Cause | Fix |
|---|---|---|
| `pnpm-deps` hash mismatch (#345) | 8ac19c9 bumped `@types/node` + lockfile without restamping the FOD hash | refresh `pnpmDepsHash` |
| `renovate-pulumi.test.ts` fails to collect | 779a418 added a second customManager matching the same file pattern; the "exactly one" guard threw at import | select by the unique `(?<versionKey>` capture group |
| `nix-output-resolve.test.ts` exits 126 | stub shebang `#!/usr/bin/env bash`; the nix sandbox has no `/usr/bin/env` | resolve `bash` and use an absolute shebang |
| `treefmt` (#347) | 7 files drifted | run the formatter |

### On #347 reporting one file

It is 7. `nix flake check` truncates a failing derivation's log to `Last 25 log lines`, which cuts the diff off mid-file. `nix log <drv>` shows the rest.

### On the 126

It reproduces **only** under `nix flake check` — the same test passes 2/2 in a dev shell, because `/usr/bin/env` exists there. Confirmed with a sandbox probe:

```
/build/tmp.5aF4Yjq5ZC/x: /usr/bin/env: bad interpreter: No such file or directory
```

## Risk

Low. The only edit outside tests and formatting is the dependency hash. No production code paths change.

## Verification

```
```

Fixes #345. Fixes #347.

### DIFF
--- a/docs/renovate/README.md
+++ b/docs/renovate/README.md
@@ -19,19 +19,19 @@ See also: [configuration](./configuration.md) (every default, group, and knob),
 The presets live in [`renovate/`](../../renovate/) at the repo root, one JSON
 file per concern, plus an aggregate that composes them:
 
-| Preset | Role |
-| ------ | ---- |
-| `default.json` | Base config: timezone, schedule, automerge policy, rate limits, labels, stability delay, digest pinning, dependency dashboard, and repo-wide `packageRules`. |
-| `minor-patch-automerge.json` | The per-ecosystem **PR groups** for minor/patch/digest updates, most with automerge enabled. |
-| `major-updates.json` | Forces every **major** update into its own non-automerged PR with a reviewer. |
-| `security.json` | Vulnerability-alert handling (automerged, labelled). |
-| `lock-maintenance.json` | Weekly lock-file maintenance. |
-| `nix.json` | Custom regex managers for Nix (`fetchPypi`, `mkHelmChartFromGitHub`, `nix run github:` pins). |
-| `pulumi.json` | Custom regex manager for annotated versions in `Pulumi.*.yaml`. |
-| `docker-images.json` | Custom regex manager for annotated container-image string literals in `.ts` sources. |
-| `yaml-manifests.json` | Custom regex managers for annotated container images and Helm chart versions in `.yaml`. |
-| `sector7-release-tarballs.json` | Updates `@jmmaloney4/sector7` GitHub release-tarball URLs in `package.json`. |
-| `all.json` | **Aggregate** — extends every preset above. Most consumers use only this. |
+| Preset                          | Role                                                                                                                                                         |
+| ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `default.json`                  | Base config: timezone, schedule, automerge policy, rate limits, labels, stability delay, digest pinning, dependency dashboard, and repo-wide `packageRules`. |
+| `minor-patch-automerge.json`    | The per-ecosystem **PR groups** for minor/patch/digest updates, most with automerge enabled.                                                                 |
+| `major-updates.json`            | Forces every **major** update into its own non-automerged PR with a reviewer.                                                                                |
+| `security.json`                 | Vulnerability-alert handling (automerged, labelled).                                                                                                         |
+| `lock-maintenance.json`         | Weekly lock-file maintenance.                                                                                                                                |
+| `nix.json`                      | Custom regex managers for Nix (`fetchPypi`, `mkHelmChartFromGitHub`, `nix run github:` pins).                                                                |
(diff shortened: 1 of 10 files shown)


jmmaloney4:
## Merge-gate audit

- **PR-Agent `/review`** — clean at open (`No major issues detected`, `No security concerns identified`); no commits pushed since, so HEAD is still the reviewed SHA `6bb550b`.
- **`mergeStateStatus: UNSTABLE`** — the only GitHub checks are 3 jobs `QUEUED` on self-hosted runners that are currently offline. Not a blocker.
- **Local validation on current HEAD** — full suite green:

```
```

  This is the first fully-green `nix flake check` on this repo since 2026-07-21.

Merging per the standing approval for the ADR 163 migration workstream.

PR #358, opened 2026-08-04
feat(onepassword): retype OnePasswordItem onto the resource plugin

jmmaloney4:
> **Stacked on #357.** Base is `fix/unbreak-flake-check`; retarget to `main` once that merges. Without it no TS check can build, so this could not be verified on its own.

## Why

Last TypeScript-side piece of the 1Password migration. The CRUD implementation landed in `provider/onepassword/item.go` in #356; this deletes the dynamic provider that was shadowing it, completing ADR 163 for this family.

`item.ts`: **622 → 125 lines**. What remains is the input surface, the secret-marking, and the alias.

## The alias is the whole cutover story

```ts
aliases: [{ type: "pulumi-nodejs:dynamic:Resource" }]
```

The engine rewrites the URN in place, so `pulumi preview` is a complete, side-effect-free dry run and **no Connect API call happens during the transition**. Keep it for at least two releases — it costs nothing and covers stacks not yet through an `up`, plus state restored from a backup taken before the retype.

## Tests: all 19 cases ported to Go

`tests/onepassword-item.test.ts` was the **only** place CRUD and Diff were exercised. The pre-existing Go tests cover pure functions (`ContentHash`, `mergedBody`, the collation guard) and never invoke Create, Update, Delete or Diff — there was no harness at all.

Deleting it without porting would have dropped that coverage silently. That matters more here than for most resources: `OnePasswordItem` adopts by title in a **shared** vault, so a wrong adoption overwrites — and can later `DELETE` — an unrelated hand-created secret.

Each Go subtest carries the original `it(...)` string **verbatim**, so the two lists diff mechanically. Suite counts corroborate the port:

| | tests | files |
|---|---|---|
| before | 301 | 25 |
| after | 282 | 24 |

— a difference of exactly the 19 cases in the 1 deleted file.

New harness (`httptest` + `kube.Fake`) serves the Connect REST API, so no Kubernetes and no Connect server are involved.

## Negative controls

A suite that passes on first write is worth distrusting, so I disabled each guard in `item.go` and confirmed the matching test fails:

| Guard disabled | Test that caught it |
|---|---|
| `if len(matches) > 1` → `if false` | `create refuses to adopt when multiple items share the title` |
| stop carrying existing keys forward | `preserves unmanaged fields/sections when adopting an item` |
| `for _, l := range priorManagedLabels` → nil | `update removes a previously-managed field dropped from input` |
| 404 tolerance → `&& false` | `delete treats a 404 as success (idempotent destroy)` |

The first control initially failed for the *wrong* reason (a fixture artifact — the per-item `GET` returned a list and it died decoding). I fixed the fixture so disabling the guard produces a real adoption, then re-confirmed.

One case is a regression guard rather than a control: `does not leak the Connect response body into thrown errors` passes trivially today because `httpx.Error` renders method/path/status only. It exists to fail if a body is ever added.

## Not in this PR

`k8s/port-forward.ts`, `DynamicResourceOptions` and `rejectCloudProviderOptions` stay until the remaining families (attic, matrix, r2) are migrated.

## Verification

```
```


jmmaloney4:
## Merge-gate audit

- **PR-Agent `/review`** — clean, and verified against the current HEAD: `(Review updated until commit .../6dc337d0f8968319e6bc7203ffdbd9664a6a4075)` matches `headRefOid`. `No major issues detected`, `No security concerns identified`.
- **PR-Agent `/improve`** — `No code suggestions found for the PR.` (The first manual run stuck in the `Work in progress ...` placeholder; a completed run superseded it.)
- **Base retargeted to `main`** now that #357 has merged. Diff is exactly the four intended files:
  ```
  packages/sector7/onepassword/index.ts
  packages/sector7/onepassword/item.ts
  packages/sector7/tests/onepassword-item.test.ts   (deleted)
  provider/onepassword/crud_test.go                 (added)
  ```
- **`mergeStateStatus: UNSTABLE`** — only queued jobs on offline self-hosted runners.
- **Local validation** — full `nix flake check --keep-going` green on this exact tree (#357's content is now `main`, so the branch I validated and the current HEAD are the same code):

```
```

### Note on the brief auto-close

This PR was auto-closed for a few minutes when #357 merged with `--delete-branch`, which removed this PR's base branch. Recovered by restoring the ref, reopening, retargeting to `main`, and re-deleting the ref. No commits or review history were lost; HEAD is unchanged at `6dc337d`. Retarget *before* merging the parent next time.

PR #359, opened 2026-08-04
fix(attic): accept an explicit kubeconfig for the admin port-forward

jmmaloney4:
## Summary

- `attic`'s `Cache` resource has the same latent RBAC gap that already caused a real incident for `litellm` (sector7#350): its admin port-forward (`kube.Transport.Connect`) always used whatever kubeconfig is ambient in the process running `pulumi up`/`refresh`, with no way for a stack to pass a narrower or different kubeconfig explicitly.
- Adds an optional, secret `CacheArgs.Kubeconfig` field (`pulumi:"kubeconfig,optional" provider:"secret"`), threaded into the `kube.Target{...}` built in `Cache.connect` (`provider/attic/cache.go`) — same tag convention and doc-comment style as `litellm.AdminTarget.Kubeconfig` and `onepassword.ItemArgs.Kubeconfig`.
- `Cache.Diff` now treats a `Kubeconfig`-only change as part of the existing `adminTarget` in-place `Update`, not a replace — mirroring `AdminTarget.Changed` in litellm. A cache replace would regenerate its signing keypair and break every client's `trustedPublicKeys`, so this must stay in-place.
- Fixed a stale doc comment on `kube.Target.Kubeconfig` that claimed "litellm and attic never pass one; only OnePasswordItem may" — no longer true after sector7#350 and this PR.

## Scope

Input plumbing only, per the litellm precedent:
- **No `__provider` state migration** — attic never went through a dynamic-provider retype, so (unlike litellm) there is no legacy state shape to migrate against. This field is additive and optional; existing state simply decodes with `Kubeconfig: ""`.
- **No TypeScript changes** — `packages/sector7/attic/admin.ts` is untouched.
- **`MintArgs` (token.go) and `TokenArgs` (token_resource.go) are untouched** — verified neither has a `.Connect` call. `MintToken` signs JWTs entirely in-process (HMAC, no cluster connectivity), so there is nothing to route a kubeconfig through. The single `.Connect` call site in the whole `provider/attic` package is `cache.go:187`, and it now carries `Kubeconfig`.

## Test plan

- [x] `provider/attic/cache_test.go`: added `TestCacheKubeconfigReachesTheTransport` — proves a set `Kubeconfig` reaches `kube.Fake.LastTarget.Kubeconfig` (plus the rest of the target), and that an unset one stays empty (ambient), mirroring litellm's `TestAdminTargetKubeconfigReachesTheTransport`.
- [x] `provider/attic/cache_test.go`: added `TestCacheKubeconfigChangeIsAnAdminTargetDiff` — proves a rotated `Kubeconfig` diffs as an in-place `adminTarget` `Update`, never a `Replace`, and that unchanged inputs stay a no-op. Mirrors litellm's `TestKubeconfigChangeIsAnAdminTargetDiff`.
- [x] `nix build .#checks.x86_64-linux.provider-go-test --no-link` — green (`provider/attic` package, all subtests).
- [x] `nix flake check --keep-going -L` — green (treefmt, pre-commit, tsc, vitest, provider-version-stamped, renovate-config; `provider-go-test` verified separately above).

## Risks

- Low. Purely additive optional field; default behaviour (empty `Kubeconfig` → ambient config) is unchanged for every existing stack. The `Diff` change only affects the *kind* of update `Kubeconfig` triggers (in-place vs. none — it was previously not a tracked field at all), not whether it changes.

## Related

- Precedent: sector7#350 (`fix(litellm): accept an explicit kubeconfig for the admin port-forward`)

### DIFF
--- a/provider/attic/cache.go
+++ b/provider/attic/cache.go
@@ -38,7 +38,22 @@ type Cache struct {
 }
 
 type CacheArgs struct {
-	Namespace string `pulumi:"namespace"`
+	// Kubeconfig is YAML; empty means the ambient default config.
+	//
+	// Without it, the admin port-forward falls back to whatever kubeconfig is
+	// ambient in the process running `pulumi up`/`refresh` — which is not
+	// necessarily the identity the rest of the stack deploys with. This is the
+	// same gap that broke litellm's admin port-forward (sector7#350):
+	//
+	//	deployments.apps "litellm" is forbidden: User "pulumi-zeus" cannot get
+	//	resource "deployments" in API group "apps" in the namespace "litellm"
+	//
+	// AdminTarget.Kubeconfig in provider/litellm/client.go added the fix there;
+	// this mirrors it for attic. Secret because a kubeconfig carries client
+	// certificates and keys — matching onepassword.ItemArgs.Kubeconfig and
+	// litellm.AdminTarget.Kubeconfig.
+	Kubeconfig string `pulumi:"kubeconfig,optional" provider:"secret"`
+	Namespace  string `pulumi:"namespace"`
 	// HS256SecretBase64 is the server's signing secret; admin tokens are minted
 	// from it locally.
 	HS256SecretBase64     string   `pulumi:"hs256SecretBase64" provider:"secret"`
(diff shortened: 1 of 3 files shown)


jmmaloney4:
## Merge-gate audit

- **Independently re-verified by me**, not just trusting the authoring agent's self-report: reverted the `Kubeconfig: a.Kubeconfig` threading line and confirmed `TestCacheKubeconfigReachesTheTransport/forwarded_when_set` fails with exactly the right symptom (`got ""`), then restored and confirmed clean.
- **`mergeStateStatus: UNSTABLE`** — only queued self-hosted-runner checks, known pattern; rebased cleanly onto main after sector7#360 merged.
- **Local validation** — full `nix flake check --keep-going` green.

Mirrors the litellm precedent (sector7#350) exactly: same field name/tag convention, same in-place-update Diff semantics (never a replace, which would regenerate the cache's signing keypair).

PR #360, opened 2026-08-04
fix(nix-output): stop diffing REPO_ROOT across machines

jmmaloney4:
## Why

Same root cause as garden#1607, different location. `NixOutput` forwards `args.repoRoot`'s resolved value into `command.local.Command`'s `environment.REPO_ROOT` — an absolute filesystem path that differs by machine/checkout. A value change there forces a full replace of the Command, cascading into everything downstream of the store path it resolves. `NixImage` composes `NixOutput` internally for its build step, so this affects **image builds** (ergon, litellm, and anything else using `nixImage`/`NixOutput`), not just Helm chart paths.

Found while verifying the onepassword retype (garden#1602) was safe to apply to ergon/prod and litellm/prod — both showed large unrelated diffs in their `NixOutput`/`NixImage`-backed resources, same shape as the gateway CRD-deletion bug garden#1607 fixed.

## Fix

Stop forwarding the resolved value into the tracked `environment` map. `nix-output-resolve.sh` now falls back to the ambient `FLAKE_ROOT` when `REPO_ROOT` isn't explicitly set. Verified first, with an isolated repro, that `command.local.Command`'s `environment` property **merges with** (doesn't replace) the ambient process environment — so this fallback actually reaches the spawned command:

```
$ echo [FLAKE_ROOT=$FLAKE_ROOT] [SET=${FLAKE_ROOT+yes}]
[FLAKE_ROOT=] [SET=]        # environment: { NIX_ATTR: "foo" } — FLAKE_ROOT still passes through
```

`args.repoRoot` is still used correctly for the synchronous, non-diffed TypeScript computations (drvPath trigger resolution, eager preview resolution) — only the diffed Command input changes.

In every known caller, `args.repoRoot`'s resolved value is already identical to the ambient `FLAKE_ROOT` at runtime (it's either `FLAKE_ROOT` directly, or an equivalent path computed via `import.meta.url`), so this is **not a behavior change** for any current call site — only the diffing goes away.

## Tests

- `nix-output.test.ts`: updated to assert `REPO_ROOT` is absent from the tracked `environment` (was asserting it was present — inverted on purpose).
- `nix-output-resolve.test.ts`: new coverage for the fallback — `FLAKE_ROOT` used when `REPO_ROOT` unset, `REPO_ROOT` preferred when both are set, clear failure when neither is set.
- **Negative control**: reverted each half of the fix independently and confirmed the corresponding new/updated test fails:
  - TS-side revert (re-add `REPO_ROOT: args.repoRoot` to the env map) → both `nix-output.test.ts` assertions fail
  - Shell-side revert (remove the `FLAKE_ROOT` fallback) → the fallback test fails with `status=1` instead of `0`

## Verification

```
✅ renovate-config   ✅ treefmt   ✅ tsc   ✅ vitest (290/290)
```

### DIFF
--- a/packages/sector7/nix-output/nix-output.ts
+++ b/packages/sector7/nix-output/nix-output.ts
@@ -6,7 +6,20 @@ import { getScriptPath } from "../scripts/index.ts";
 export interface NixOutputArgs {
 	/** Flake attribute path (e.g. "packages.x86_64-linux.lens-api-image") */
 	nixAttr: pulumi.Input<string>;
-	/** Absolute path to the repo root containing the flake. */
+	/**
+	 * Absolute path to the repo root containing the flake.
+	 *
+	 * Used locally (drvPath trigger resolution, eager preview resolution) —
+	 * never forwarded into the spawned command's tracked `environment` input.
+	 * That command instead reads REPO_ROOT from its own ambient environment
+	 * at execution time, falling back to FLAKE_ROOT (see
+	 * nix-output-resolve.sh). Forwarding this value would bake an absolute,
+	 * machine-specific filesystem path into a diffed Pulumi input, forcing a
+	 * spurious replace whenever the same stack is applied from a different
+	 * checkout path than whoever last applied it. In every known caller this
+	 * value is already identical to the ambient FLAKE_ROOT at runtime, so the
+	 * command sees the same path either way — it just isn't tracked.
+	 */
 	repoRoot: pulumi.Input<string>;
 	/**
 	 * Select a named output from a multi-output nix derivation.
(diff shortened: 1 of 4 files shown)


jmmaloney4:
Re: **Ambient FLAKE_ROOT dependency** — non-blocking, documenting the rationale.

This isn't a new failure mode. Every current caller constructs `repoRoot` from `process.env.FLAKE_ROOT` (or an equivalent) *before* reaching `NixOutput` — e.g. `deploy/platform/gateway/index.ts`:

```ts
const flakeRoot = process.env.FLAKE_ROOT;
if (!flakeRoot) {
  throw new Error("Set FLAKE_ROOT (the devshell sets this automatically).");
}
```

So an automation context that never exported `FLAKE_ROOT` was *already* failing loudly before this PR — just one call frame earlier, with a different error message. This PR moves the same requirement from "the caller checks it" to "the script checks it" (`nix-output-resolve.sh`'s `Required env var REPO_ROOT is not set`), with the same net effect: no `FLAKE_ROOT` in the environment → loud, immediate failure either way, never a silent wrong build.

Pushed a separate fix (984899f) for the actual gap this line of investigation surfaced: nothing previously *enforced* that a caller's `repoRoot` matches the ambient `FLAKE_ROOT` if they ever diverged. Added a `pulumi.log.warn` for exactly that case.

jmmaloney4:
## Merge-gate audit

- **PR-Agent `/review`** — converged at current HEAD `984899f` (`(Review updated until commit .../984899f...)`), clean: `No major issues detected`.
- **PR-Agent `/improve`** — manually re-triggered after the persistent comment's marker went stale (documented pattern); fresh run on current HEAD: `No code suggestions found for the PR.`
- **One `/review` finding addressed as non-blocking with rationale** (posted in-thread): the "ambient FLAKE_ROOT dependency" observation isn't a new failure mode — every caller already required `FLAKE_ROOT` before reaching `NixOutput`; this just moves the same check one frame later, with a clearer error message.
- **One `/review` finding fixed properly**: added a `pulumi.log.warn` guard for the case the review actually flagged (repoRoot silently diverging from ambient FLAKE_ROOT) — no runtime enforcement existed before, only a doc comment.
- **Both `/improve` suggestions applied**: moved the warn-spy assertion/restore into try/finally in the new tests; passed `this` to `pulumi.log.warn` for resource association in the Pulumi UI.
- **Local validation** — full `nix flake check --keep-going` green throughout, including 3 rounds of negative-control verification (TS-side revert, shell-side revert, warning-guard revert) all failing the correct test.

PR #361, opened 2026-08-04
fix(nix-output,nix-image): pipe scripts via stdin instead of a resolved path

jmmaloney4:
## Why

A second, separate machine-path-drift bug in the same family as #360 and garden#1607 — found while previewing `ergon/prod` immediately after #360 merged, to confirm the fix worked. It didn't clean the diff up.

`getScriptPath()` resolves an absolute path through `node_modules` (`__dirname` + filename). That path is baked directly into the tracked `create` command string on three `command.local.Command` resources: `NixOutput`'s resolve command, and `NixImage`'s `resolveCmd`/`pushCmd`.

This is actually **worse** than the `REPO_ROOT` bug #360 fixed: the path varies not only by checkout location, but on **every single sector7 version bump** — pnpm encodes the resolved package version into the `.pnpm` virtual-store directory name (`.pnpm/@jmmaloney4+sector7@<url>_<hash>/`). So every release, regardless of whether it touched these scripts at all, forced a replace of every `NixOutput`/`NixImage`-backed resource on the next apply — an unnecessary image rebuild and repush for every consumer (ergon, litellm, ...), every time sector7 bumps.

## Fix

Read the script's own content at construction time (a local, synchronous, non-diffed read) and pipe it via the Command's `stdin` property, with a fixed `create: "bash -s"` instead of the resolved path. The tracked input now depends on **what the script does**, not **where it happens to live on disk** — a real diff only fires when the script's content genuinely changes.

## Verification

End-to-end isolated repro: applied a `command.local.Command` using this pattern from one absolute path (simulating `__dirname`), then previewed the identical program from a completely different path:

```
Resources:
    2 unchanged
```

Negative control: reverted to the path-based `create` and confirmed the "create stays a fixed string" assertion in `nix-output.test.ts` fails, with the exact resolved path leaking through.

## Tests

Updated both test files' mock detection — previously keyed on `create.includes("<script>.sh")`, which no longer distinguishes anything since `create` is now always `"bash -s"`. Switched to matching distinctive content in `stdin`. Added explicit assertions that `create` is the fixed string and `stdin` carries the expected script content.

```
✅ renovate-config   ✅ treefmt   ✅ tsc   ✅ vitest (35/35 targeted, full suite green)
```

### DIFF
--- a/packages/sector7/nix-image/nix-image.ts
+++ b/packages/sector7/nix-image/nix-image.ts
@@ -1,3 +1,4 @@
+import { readFileSync } from "node:fs";
 import * as command from "@pulumi/command";
 import * as pulumi from "@pulumi/pulumi";
 import { NixOutput } from "../nix-output/nix-output.ts";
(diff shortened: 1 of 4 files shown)


jmmaloney4:
## Merge-gate audit

- **Local validation** — full `nix flake check --keep-going` green.
- **Verified end-to-end**, not just unit tests: applied an isolated repro from one absolute path, previewed the identical program from a completely different path, confirmed `2 unchanged`. Negative control confirmed reverting to the path-based `create` reintroduces the drift.

PR #362, opened 2026-08-05
feat(matrix): add __provider state migration for Room/BotAccount

jmmaloney4:
## Summary

Adds the `__provider` state migration for `matrix.Room` and `matrix.BotAccount` — the prerequisite for retyping `MatrixRoom`/`MatrixBotAccount` off the garden-local TypeScript dynamic providers (`deploy/services/matrix/matrix-room.ts`, `matrix-bot-account.ts`) onto this Go resource plugin, per ADR 163.

`infer` treats an unrecognized property in stored state as a hard decode **failure**, not something silently ignored. Legacy state written by the dynamic provider carries a `__provider` field (the serialized JS closure — the very thing ADR 163 exists to stop storing), which `RoomState`/`BotAccountState` don't declare. Without a migration, the upcoming alias-based retype would fail preview outright:

```
Unrecognized field '__provider' on 'matrix.RoomState'
Unrecognized field '__provider' on 'matrix.BotAccountState'
```

instead of diffing to nothing as intended.

This follows the exact pattern already used by `d1` and `onepassword`: a `roomStateV0`/`botAccountStateV0` struct per resource that embeds the current State plus a *required* `__provider` field, dropped by an identity-copy migration in `StateMigrations`.

### The gotcha this guards against

`infer` tries a migrator's old shape *before* the normal decode, and only falls through to the normal decode when that old-shape decode **fails**. So `__provider` must be tagged `pulumi:"__provider"` — **not** `,optional`. Tagging it optional makes the old shape a strict superset that *also* matches already-migrated, plugin-native state, so the migration would fire on every future read forever instead of only on legacy dynamic-provider state. This exact mistake was made and caught by review during the d1 migration (sector7#354).

`TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState` guards this for both `Room` and `BotAccount`. I verified it's load-bearing by temporarily flipping both tags to `,optional` locally, confirming the guard test fails for both subtests, and flipping them back.

## Test plan

- [x] `go test ./...` in `provider/` — full suite green, including the new `provider/matrix/migrations_test.go`.
- [x] `nix build .#checks.x86_64-linux.provider-go-test --no-link -L` — green.
- [x] Guard test verified to actually catch the `,optional` mistake (flipped tag → test failed with the expected message → flipped back → green).
- [x] Real gRPC path exercised via `integration.NewServer` + `srv.Diff` (not just Go-level unit calls), for both resources:
  - legacy state (with `__provider`, modeled on the actual TypeScript dynamic-provider outputs) decodes and diffs to **zero changes** against freshly-retyped inputs;
  - a genuinely changed field (`name` for Room, `password` for BotAccount) still plans an update — the migration must not swallow real changes;
  - state with **no** `__provider` decodes fine via the normal path, unaffected (additive-safety).
- [x] `nix fmt` on the new files — no changes needed.
- [ ] `nix flake check --keep-going` (full repo) — running; will report result.

## Scope

The TypeScript retype of `MatrixRoom`/`MatrixBotAccount` onto the `pulumi-nodejs:dynamic:Resource` alias (the garden-local repo side) is explicitly **out of scope** for this PR — this only adds the Go-side migration that retype's `pulumi preview` will depend on.

## Risks

None to existing resources — this is a new, self-contained file (`provider/matrix/migrations.go` + its test) with no central registration point and no changes to `provider.go` or any other resource family.

### DIFF
--- a/provider/matrix/migrations.go
+++ b/provider/matrix/migrations.go
@@ -1,0 +1,87 @@
+package matrix
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi-go-provider/infer"
+)
+
+// State migrations off the dynamic provider.
+//
+// Room and BotAccount replace garden-local TypeScript dynamic providers
+// (deploy/services/matrix/matrix-room.ts, matrix-bot-account.ts — see the
+// package doc comment on botaccount.go). Live state written by either one
+// carries `__provider`, the serialized JavaScript closure this plugin exists
+// to stop storing. infer treats an unrecognized property as a decode FAILURE
+// rather than something quietly ignored:
+//
+//	Unrecognized field '__provider' on 'matrix.RoomState'
+//	Unrecognized field '__provider' on 'matrix.BotAccountState'
+//
+// so without a migration the alias retype does not diff to nothing; it fails
+// outright at the first preview. Same shape and same mechanism as d1 and
+// onepassword; see provider/d1/migrations.go for the most heavily-commented
+// version of this pattern.
+//
+// One shape suffices for each resource. Neither RoomState nor BotAccountState
+// has a sentinel-encoded optional the way litellm's maxBudget did — every
+// property is a plain string, []string or *bool, so an empty/nil value is a
+// valid value of its own type and decodes fine whichever path it takes.
+// __provider is the only obstacle. (RoomArgs.IsDirect is a *bool rather than
+// bool specifically so a room with no isDirect in its inputs decodes as nil
+// instead of a false that Diff would then treat as a real value — see
+// boolPtrEqual in room.go — but that is a property of RoomState itself, not
+// something this migration needs to work around.)
+//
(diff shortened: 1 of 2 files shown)


jmmaloney4:
## Merge-gate audit

- **PR-Agent `/review`** — clean first-run, no SHA marker but no commits since open: `No major issues detected`, `No security concerns identified`.
- **Independently re-verified by me**: flipped both `roomStateV0.Provider`/`botAccountStateV0.Provider` tags to `,optional` and confirmed `TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState` fails for both, with the exact right message.
- **`mergeStateStatus: UNSTABLE`** — only queued self-hosted-runner checks.
- Full `nix flake check --keep-going` green.

Go-side of the matrix family. TS retype (garden-local, `deploy/services/matrix/matrix-room.ts`/`matrix-bot-account.ts`) follows separately.

PR #363, opened 2026-08-05
feat(r2): add __provider state migration for Object

jmmaloney4:
## Summary

Adds the `__provider` state migration for `r2.Object` — the next resource in ADR 163's dynamic-provider-to-Go-plugin cutover, after litellm, d1, and onepassword. The Go CRUD implementation for `Object`/`ObjectState` already exists in `provider/r2/object.go`; this PR adds only the migration the alias-based TypeScript retype will need once it lands (that retype is out of scope here — see below).

## Why

`infer` treats an unrecognized property in stored state as a hard decode **failure**, not something silently ignored. Legacy state written by the TypeScript dynamic provider (`packages/sector7/r2/r2object.ts`) carries a `__provider` field (the serialized JS closure) that `ObjectState` doesn't declare. Without a migration, the alias retype does not diff to nothing at preview time — it fails outright:

```
Unrecognized field '__provider' on 'r2.ObjectState'
```

## The load-bearing gotcha

`infer` tries each registered migration's "old" shape *before* falling through to the normal decode, on **every** read/diff, not just ones touching legacy state. So `objectStateV0.Provider` is tagged plain `` pulumi:"__provider" `` — **not** `,optional`. Tagging it optional would make the old shape a superset that also matches already-migrated, plugin-native state, so the migration would fire on every single future read instead of only on legacy dynamic-provider state.

This exact mistake was made and caught by review during the d1 migration (sector7#354). A guard test, `TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState`, pins the tag so it can't regress silently — verified locally by temporarily flipping the tag to `,optional` and confirming the guard test fails, then reverting.

## Scope

Only `R2Object`/`ObjectState`. `provider/r2/zonecachepurge.go`'s `ZoneCachePurge` has no live TypeScript dynamic-resource usage anywhere in `jmmaloney4/garden` (verified by grep across the repo), so there is no legacy state to migrate for it — no migration was added there. The TypeScript retype of `packages/sector7/r2/r2object.ts` / `packages/sector7/r2/index.ts` into a thin `pulumi.CustomResource` wrapper with a `pulumi-nodejs:dynamic:Resource` alias is explicitly out of scope for this PR.

## Test plan

- [x] `provider/r2/migrations_test.go` exercises the real gRPC path via `integration.NewServer` + `srv.Diff` (not just Go-level unit calls):
  - `TestDiffAcceptsLiveDynamicProviderState` — live dynamic-provider state (with `__provider`) diffs to nothing against the retyped wrapper's inputs.
  - `TestRealChangeSurvivesTheMigration` — a genuine input change (`key`) still plans an `UpdateReplace`, proving the migration doesn't swallow real diffs.
  - `TestPluginNativeStateStillDecodes` — state with no `__provider` decodes via the normal path unaffected.
  - `TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState` — structural guard on the tag; manually verified it fails when the tag is flipped to `,optional`.
- [x] `go test ./...` — full provider module green (all packages, including r2, d1, onepassword, matrix, litellm, attic).
- [x] `nix build .#checks.x86_64-linux.provider-go-test --no-link` — green.
- [x] `nix flake check --keep-going` (x86_64-linux) — all 7 applicable checks green: `provider-go-test`, `renovate-config`, `pre-commit`, `provider-version-stamped`, `tsc`, `treefmt`, `vitest`. (aarch64-darwin/aarch64-linux/x86_64-darwin omitted — cross-platform, not buildable from this runner.)

## Risks

Low. This is an additive-only Go file (`provider/r2/migrations.go`) plus its test — no changes to `object.go`'s CRUD logic, no changes to any TypeScript, no central registration required (`StateMigrations` is picked up structurally via `infer.CustomStateMigrations[ObjectState]`). Does not itself trigger the TypeScript cutover; that's a separate follow-up PR.

### DIFF
--- a/provider/r2/migrations.go
+++ b/provider/r2/migrations.go
@@ -1,0 +1,59 @@
+package r2
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi-go-provider/infer"
+)
+
+// State migration off the dynamic provider.
+//
+// R2Object replaces the garden-local TypeScript dynamic provider
+// (packages/sector7/r2/r2object.ts, deployed today via
+// deploy/www/theoreticaledge.com's uploadStaticAssets/R2Object pipeline). The
+// dynamic provider's create/update handlers wrote state as `{ ...inputs, etag
+// }`, and the Pulumi dynamic-resource runtime additionally persists
+// `__provider`, the serialized JavaScript closure this plugin exists to stop
+// storing. infer treats an unrecognized property as a decode FAILURE rather
+// than something quietly ignored:
+//
+//	Unrecognized field '__provider' on 'r2.ObjectState'
+//
+// so without this migration the alias retype does not diff to nothing; it
+// fails outright at the first preview. Same shape and same mechanism as d1
+// and onepassword; see provider/d1/migrations.go for the most
+// heavily-commented version of this pattern.
+//
+// One shape suffices. There is no sentinel-encoded optional the way
+// litellm's maxBudget was — every ObjectArgs/ObjectState property is a plain
+// string, so an empty value is a valid value of its own type and decodes
+// fine whichever path it takes. __provider is the only obstacle.
+//
+// The migration is additive-safe: state that never passed through the
+// dynamic provider has no __provider, fails this migrator's decode ("doesn't
+// fit into the migrator", per infer's own docs), and falls through to the
+// normal path unchanged. Verified against the real gRPC path in
+// migrations_test.go using a shape modeled on r2object.ts's stateOuts.
+type objectStateV0 struct {
+	ObjectState
+	// Provider is the serialized dynamic-provider closure. Declared only so
+	// it can be dropped — nothing reads it.
+	//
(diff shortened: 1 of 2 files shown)


jmmaloney4:
## Merge-gate audit

- **PR-Agent `/review`** — clean: `No major issues detected`, `No security concerns identified`.
- **Independently re-verified by me**: flipped `objectStateV0.Provider`'s tag to `,optional` and confirmed `TestProviderTagIsRequiredSoTheMigrationOnlyMatchesLegacyState` fails with the exact right message.
- **`mergeStateStatus: UNSTABLE`** — only queued self-hosted-runner checks.
- Full `nix flake check --keep-going` green.

Go-side of the r2 family. TS retype (`packages/sector7/r2/r2object.ts`, `r2/index.ts`) follows separately — 1 live resource (theoreticaledge.com).

PR #364, opened 2026-08-05
feat(r2): retype R2Object onto the resource plugin

jmmaloney4:
## Why

Last dynamic-provider resource in this package. The Go CRUD (AWS Sig V4 signing, MD5-based diff, replace-on-key/bucket/account change) and `__provider` migration landed in sector7#363; this deletes the JS dynamic provider that was shadowing it, completing ADR 163 for the r2 family.

## What

`r2object.ts`: **1036 → 678 lines**. Removed entirely: the native AWS Sig V4 signing implementation (`SigV4Credentials`, `deriveSigningKey`, `signedFetch`, `rfc3986Encode`), `uploadObjectToR2`/`deleteObjectFromR2`, the `r2ObjectProvider` `dynamic.ResourceProvider`, and `R2Object`'s own `rejectCloudProviderOptions` call — no longer meaningful, since `R2Object` is now plugin-backed and `provider`/`providers` are ordinary supported options (same as onepassword/matrix).

`ZoneCachePurge`, `uploadAssets` and `uploadStaticAssets` are **untouched** — `ZoneCachePurge` stays a dynamic resource (out of scope, no live TypeScript usage anywhere in `jmmaloney4/garden`), and the two upload helpers are orchestration around `R2Object`, not part of its own CRUD.

## Tests

Rewrote `r2object.test.ts`. The old "R2Object provider" describe block tested the dynamic provider's `check`/`diff` logic directly — that coverage now lives in `provider/r2/object_test.go` (already existed before this PR) and is superseded, not lost.

Added coverage for what the TS wrapper actually owns:
- registers under `sector7:r2:Object` with the dynamic-provider alias
- accepts `provider`/`providers` instead of throwing
- marks both credential fields secret on the input side — verified via a `vi.fn()` spy on `pulumi.secret`. An earlier identity-mock version of this test passed whether or not the constructor called `secret()` at all; negative-control testing caught that it wasn't actually asserting anything, which is why the spy version exists instead.

`zone-cache-purge.test.ts` needed a matching `CustomResource` stub added to its own `@pulumi/pulumi` mock: it only imports `purgeZoneCache` from this same module, but `class R2Object extends pulumi.CustomResource` is evaluated at module-load time regardless of which export a test file actually uses — its mock needed the export defined even though that file never constructs an `R2Object`.

## Verification

```
```

### DIFF
--- a/packages/sector7/r2/r2object.ts
+++ b/packages/sector7/r2/r2object.ts
@@ -7,7 +7,6 @@ import {
 	type CustomResourceOptions,
 	dynamic,
 	type Input,
-	type Output,
 	type Resource,
 } from "@pulumi/pulumi";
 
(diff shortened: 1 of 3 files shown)


jmmaloney4:
## Merge-gate audit

- **PR-Agent `/improve`** — never produced an actionable suggestion at any point on this PR; the manual re-trigger got stuck in the `Work in progress ...` placeholder rather than completing. Per the documented stuck-placeholder exception: `/review` is clean and current, there are no stale suggestions from an earlier SHA to verify against (none ever rendered), and local validation is green — treating this as converged.
- **Local validation** — full `nix flake check --keep-going` green, including the negative-control-verified test suite for the retype.

PR #365, opened 2026-08-07
feat(attic): retype Cache/Token onto the resource plugin

jmmaloney4:
## Why

Completes ADR 163: the last remaining sector7 dynamic provider. `AtticCache`/`AtticToken` move from Pulumi *dynamic* providers (which serialize their JS closure into stack state and re-execute the stored copy on refresh/delete — the root cause of a prior production incident) onto the Go resource plugin.

## What changed (scope grew from the original Go-only migration)

1. **State migrations** (`provider/attic/migrations.go`): drop the legacy `__provider` field from decoded state — same shape as d1/onepassword/matrix.
2. **Token-collision fix, found during validation** (`Cache.Annotate` / `Token.Annotate` in `cache.go` / `token_resource.go`): the plugin now registers as `sector7:atticprovider:{Cache,Token}` instead of the package-derived `sector7:attic:{Cache,Token}`.
3. **TypeScript retype** (`packages/sector7/attic/admin.ts`): `AtticCache`/`AtticToken` become thin `CustomResource` wrappers, aliased to the old nested `dynamic.Resource` child's exact URN (type, name suffix, and parent).

## Why the token had to change (#2)

Attic's old structure is architecturally different from every other migrated family: `AtticCache`/`AtticToken` were `ComponentResource`s (type `sector7:attic:Cache`/`Token`) wrapping an inner `dynamic.Resource` CHILD that held the real state — not a single flat dynamic resource like matrix/r2/onepassword/litellm/d1.

The Go plugin's package-derived token would have been `sector7:attic:Cache`/`Token` — the *exact same string* the old `ComponentResource` already used live. I verified empirically (a throwaway Go plugin built against a real nested old state, not just reasoned about) that this collision is dangerous: when the new resource's own identity coincidentally matches the old component's URN, the Pulumi engine treats that as a **direct match and never consults the declared alias at all**. The real child (holding the actual secrets/id) gets silently discarded and replaced from the component's near-empty bookkeeping state instead of being adopted.

With the collision removed (`sector7:atticprovider:*`), the alias is the only candidate; the child adopts cleanly as an in-place `update` (verified via a real `pulumi up` against the repro state — the adopted resource's `created` timestamp and `id` are preserved, not regenerated). The old component (`custom: false`, no `id`) drops out of state as harmless bookkeeping cleanup — no provider `Delete` RPC is ever issued for it.

## Fixtures

Modeled on real `attic/prod` state (`pulumi stack export -C deploy/services/attic -s prod`), field names and shapes read off the actual stack rather than imagined.

## Verification

- Negative control on both `Cache` and `Token` migrations: flipping the `__provider` tag to `,optional` fails the guard test with the exact expected message.
- Negative control on the token-collision fix: reverting to `sector7:attic:{Cache,Token}` fails the new construction tests with the expected mismatch.
- Full nested-alias mechanism verified end-to-end against a real local Pulumi backend: built a throwaway Go plugin mirroring the real fix, deployed a faithful old component+child structure, then previewed and applied the new flat-resource program against it — clean `update`, id/creation-timestamp preserved, no data loss.
- `go test ./...` (all packages), `pnpm vitest run` (278 tests), `tsc --noEmit`, `biome check` all pass on the full tree.
- `nix build .#checks.provider-go-test` passes against the staged tree (the untracked-files gap from the original migration commit does not recur here — everything was staged before running).

**Caught and fixed a real process error along the way** (original migration commit): the new files were never `git add`ed, so nix's flake-check evaluation was silently running against the *old* source tree for every check up to that point, including an apparently-clean first negative-control run that had not actually exercised anything. Staged the files and reran both the positive test and the negative control from scratch before trusting either.

### DIFF
--- a/packages/sector7/attic/admin.ts
+++ b/packages/sector7/attic/admin.ts
@@ -1,735 +1,78 @@
 import * as pulumi from "@pulumi/pulumi";
-import {
-	type CustomResourceOptions,
-	dynamic,
-	type Output,
-} from "@pulumi/pulumi";
-import { withPortForward } from "../k8s/port-forward.ts";
 import type { AtticCacheArgs, AtticTokenArgs } from "./config-types.ts";
-import {
-	type AtticCacheGrants,
-	type AtticCachePermissionFlags,
-	mintAtticToken,
-	parseDurationSeconds,
-} from "./token.ts";
+import { parseDurationSeconds } from "./token.ts";
 
 /**
- * Resource options accepted by sector7 dynamic resources.
+ * An Attic binary cache, administered through the sector7 resource plugin.
  *
- * Pulumi dynamic resources are executed by the Node.js dynamic provider runtime,
- * not by a cloud provider plugin. Passing `provider` or `providers` makes Pulumi
- * route the resource through the wrong provider bridge, which fails with a
- * misleading `pulumi-nodejs:dynamic:Resource` unknown-token error.
- */
-export type DynamicResourceOptions = Omit<
-	CustomResourceOptions,
-	"provider" | "providers"
->;
-
-// ---------------------------------------------------------------------------
-// Resolved provider input/state shapes
-//
-// Pulumi resolves every `Input<T>` before invoking the dynamic provider, so the
-// callbacks below see plain JSON values only. Optional fields are normalized to
-// sentinel empty values ("" / [] / {}) by the wrapping component so the provider
-// never has to reason about `undefined`.
-// ---------------------------------------------------------------------------
-
-/** Coordinates + signing secret for reaching the Attic cache-config API. */
-interface AdminTarget {
-	namespace: string;
-	hs256SecretBase64: string;
-	deploymentName: string;
-	port: number;
-}
-
-interface CacheProviderInputs extends AdminTarget {
-	cacheName: string;
-	isPublic: boolean;
-	priority: number;
-	storeDir: string;
-	upstreamCacheKeyNames: string[];
-	/** "" when unset (server default / `Global`). */
-	retentionPeriodSeconds: number | "";
-}
-
-interface CacheProviderState extends CacheProviderInputs {
-	/** Public NAR-signing key, read back from the cache config. */
-	publicKey: string;
-}
-
-interface TokenProviderInputs {
-	hs256SecretBase64: string;
-	sub: string;
-	validitySeconds: number;
-	caches: AtticCacheGrants;
-}
-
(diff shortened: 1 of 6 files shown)


jmmaloney4:
Re: the reviewer guide's secret-output concern on \`AtticToken.token\` — verified, false positive.

\`token_resource.go\` (unchanged by this PR, so not visible in the diff) already tags the field:

\`\`\`go
type TokenState struct {
	TokenArgs
	Token     string \`pulumi:"token" provider:\"secret\"\`
	...
\`\`\`

Confirmed against a fresh build of this branch's plugin — \`pulumi package get-schema\` shows:

\`\`\`json
"token": {
  "type": "string",
  "secret": true
}
\`\`\`

\`provider:"secret"\` on a Go-schema output is the correct, idiomatic mechanism (this is how every real, non-dynamic Pulumi provider marks secret outputs) — the engine honors it automatically at RegisterResource time, no TS-side \`pulumi.secret()\`/\`additionalSecretOutputs\` needed. Those were only necessary on the OLD dynamic-provider path because dynamic resources have no declarative schema to carry that annotation.

jmmaloney4:
## Merge-gate audit

- \`go build ./...\` + \`go test ./...\` (all packages): pass.
- \`pnpm vitest run\` (278 tests, full repo), \`tsc --noEmit\`, \`biome check\`: pass.
- \`nix build .#checks.provider-go-test\` against the fully-staged tree: pass.
- Negative control (migration): flipping \`__provider\` to \`,optional\` fails the guard test with the exact expected message.
- Negative control (token fix): reverting \`sector7:atticprovider:*\` back to \`sector7:attic:*\` fails the new construction tests with the expected type mismatch.
- End-to-end mechanism verified against a real local Pulumi backend (not just unit tests): built a throwaway Go plugin mirroring the exact fix, deployed a faithful old ComponentResource+dynamic.Resource child structure, then previewed AND applied (\`pulumi up\`) the new flat-resource program against it. Result: clean \`update\` (not replace), \`id\` and \`created\` timestamp preserved on the adopted resource, old component dropped from state with no orphaned/duplicate entries.
- PR-Agent \`/review\`: one concern raised (AtticToken.token secret-output handling) — verified false positive, evidence posted above (schema shows \`"token": {"secret": true}\`).
- PR-Agent \`/improve\`: no suggestions.

Non-blocking checks still queued (\`add-to-project\`, \`adr-management placeholder push\`) — same bookkeeping workflows unrelated to code correctness seen on every PR this session.

Merging.

PR #372, opened 2026-08-13
chore(nix): bump jackpkgs to ADR-047 captured-symlink fix

jmmaloney4:
## What the bug was

`jackpkgs.nodejs` captures a pnpm workspace's dependency tree into a Nix derivation, but it copies **only** `node_modules` subtrees — never the workspace package sources. pnpm materialises workspace packages as *relative* symlinks, so inside the captured `$out` those links either dangled or resolved to a **skeleton directory** that existed only because it hosted a nested `node_modules/` copy. `dontCheckForBrokenSymlinks = true` had been set upstream to silence the resulting `noBrokenSymlinks` failure, which shipped the broken links rather than fixing them. At runtime `mkWorkspaceRuntime` links those poisoned per-package trees into a live checkout, and Node's ESM resolver finds the poisoned link before walking up to the correct root-level workspace symlink — `ERR_MODULE_NOT_FOUND`.

Upstream fix: **jackpkgs#367 / ADR-047** (`docs/internal/designs/047-strip-workspace-symlinks-from-captured-node-modules.md`), commit `5b6b6f27c537429d1d2be2a016fe87dcfe656ccc`. A shared `jackpkgsLib.nodejs.captureNodeModules` helper now strips every symlink in `$out` that does not resolve to a real path inside a `node_modules` subtree of `$out`; dependency-store links (`-> .pnpm/...`, `.bin` entries) all resolve inside one and survive. `dontCheckForBrokenSymlinks` was removed, so `noBrokenSymlinks` is the regression guard again.

## Measurement in this repo

Captured store path obtained from the devshell:

```
nix develop --command bash -c 'echo "$PATH"' | tr ":" "\n" \
  | grep -m1 "node_modules/.bin$" | sed "s|/node_modules/.bin$||"
```

Classification is exactly the ADR-047 strip rule: a symlink is legitimate iff it resolves inside `$NM/node_modules/…` or `$NM/*/node_modules/…`.

| | jackpkgs `496252f8` (before) | jackpkgs `250cd0bd` (after) |
|---|---|---|
| captured store path | `…-gd11jdqxx0327brb6r84abjn8qnz4akg-node_modules` | `…-r966ks7a5w9dp99c7vi2332c9z7pl7x3-node_modules` |
| total symlinks | 957 | 956 |
| **DANGLING** | **0** | **0** |
| **FOREIGN** | **1** | **0** |

The single foreign link before the bump:

```
FOREIGN <out>/node_modules/.pnpm/node_modules/@jmmaloney4/sector7
     -> <out>/packages/sector7
```

and `<out>/packages/sector7` was the skeleton ADR-047 describes — it contained nothing but `node_modules/`, no `package.json`, no source:

```
$ ls -la <out>/packages/sector7/
dr-xr-xr-x  node_modules
```

After the bump that entry is gone (`.pnpm/node_modules/@jmmaloney4/` is now an empty directory), the captured tree builds with `noBrokenSymlinks` **enabled**, and exactly one symlink disappeared — the poisoned one. Nothing legitimate was stripped.

The regression guard flipped between the two pins, verified in `modules/flake-parts/nodejs.nix`:

| | `496252f8` | `250cd0bd` |
|---|---|---|
| `dontCheckForBrokenSymlinks` | `= true` (line 116) | removed |
| `installPhase` | inline copy | `jackpkgsLib.nodejs.captureNodeModules` |

So the "after" tree is not merely clean by inspection — `noBrokenSymlinks` ran against it during the build and passed.

sector7 has a single workspace package with no `workspace:*` inter-package dependencies, so it carried one instance of the bug rather than the wider fan-out seen in `garden`. It is the same defect, and it is gone.

## Validation

`nix flake check -L` (x86_64-linux) — **all green**:

| check | result |
|---|---|
| `vitest` | ✅ 278 tests, 24 files |
| `tsc` | ✅ |
| `treefmt` | ✅ |
| `pre-commit` | ✅ |
| `renovate-config` | ✅ |
| `provider-version-stamped` | ✅ |
| `provider-go-test` | ✅ (`go test -race ./...`) |

`nix develop --command true` — enters cleanly.

### A note on CI coverage

`nix flake check` does **not** run in this repo's CI: the `❄️ nix (dogfood)` workflow is `disabled_manually` (last runs 2026-05-20, all failing on `main`), and `📦 pnpm (dogfood)` only triggers on push to `main`/tags, not on pull requests. So the checks table above is the *only* place the nix gate ran for this change — it is local (`x86_64-linux`), not CI-attested. Flagging that explicitly rather than letting a green PR page imply coverage that isn't there. Re-enabling that workflow is out of scope here.

## Repo-side fixes

This is a 166-commit jackpkgs jump (pinned since 2026-05-28), so some drift was expected. All three fixes are mechanical:

1. **`pnpmDepsHash`** updated to `sha256-LeFkdfR+2ilF+O2PnAHUY9vjAdn3JB41RIifDabMxao=`. Three changes in `modules/flake-parts/nodejs.nix` between the pins each independently alter the fixed-output deps derivation: pnpm 11.1.1 -> 11.11.0, `nodejs-slim` now overridden to `nodejs-slim_latest` (NixOS/nixpkgs#525627 workaround), and `fetcherVersion` **3 -> 4**. The hash was re-derived from a real fetch (placeholder hash -> read the `got:` line), not carried over.
2. **`jackpkgs.pre-commit.adr.directory = "docs/internal/designs"`.** The `adr-conflict-check` hook is new since the old pin and defaults to `docs/internal/decisions`, a path this repo has never used — it failed with `directory not found`. sector7's ADRs are `000-adr-template.md` … `034-workersite-static-assets.md`, contiguous, and pass once the hook is pointed at them.
3. **Three ADR files reformatted.** The newer `mdformat` escapes a bare `$` in prose (`$0.36` → `\$0.36`); 10 such escapes across `005`, `020`, `030`. Escaping and reflow only — normalising `\$`→`$` and collapsing whitespace makes all three byte-identical to their pre-bump versions.

   Follow-up in `fbc0997`: mdformat escaped only the *third* `$` on `020` line 200, leaving `($5/mo)` and `$0.75/GB-month` bare. Those two pair up under a math-aware renderer (MyST dollarmath / GitHub `$…$`) and swallow the middle of the sentence into a math span. Rendering every ADR whole-file finds exactly one such span in the corpus — this line — both before and after the bump; escaping the two remaining delimiters takes it to zero. Pre-existing defect, fixed here because the bump already modifies that exact line.

Release/versioning machinery was not touched.

## Lockfile scope

Only `jackpkgs` and its transitive inputs moved. This repo declares `nixpkgs`, `flake-parts` and `systems` as `follows` of `jackpkgs`, so it holds no independent pins — every other line in the `flake.lock` diff is a jackpkgs-owned input moving with it (nixpkgs `d233902` → `6cdc7fc`, pre-commit-hooks, treefmt-nix, fenix, uv2nix, pyproject-nix, home-manager, nix-unit, bun2nix). `250cd0bd` is 2 commits ahead of the ADR-047 fix commit and 0 behind it.


jmmaloney4:
## Independent review — verified, no blocking findings

Stage 2 review of `07c6aa4`. Everything below was re-derived from scratch in a fresh checkout, not taken from the authoring report.

### 1. Lockfile scope

| claim | verdict |
|---|---|
| new pin contains the ADR-047 fix | ✅ GitHub compare `5b6b6f2…250cd0b` → `ahead_by: 2, behind_by: 0` — the fix is an ancestor |
| no unrelated input moved | ✅ node **set** is byte-identical between `origin/main` and HEAD; no `owner`/`repo`/`ref`/`url` field changed on any node — only `rev`/`narHash`/`lastModified` |
| every moved input is jackpkgs-owned | ✅ root `inputs` are `{jackpkgs, nixpkgs→jackpkgs/nixpkgs, flake-parts→jackpkgs/flake-parts, systems→jackpkgs/systems}` — `jackpkgs` is the only independently-declared pin |

One structural change beyond revs: `pre-commit-hooks` dropped its `gitignore` follows-edge (upstream removed that input). The `gitignore` node itself is still present and still consumed elsewhere in the jackpkgs tree. Benign.

### 2. The ADR-047 measurement, re-run independently

Both store paths were still resident, so I re-classified **both** rather than trusting the reported numbers.

```
before  /nix/store/gd11jdqxx0327brb6r84abjn8qnz4akg-node_modules   957 links   DANGLING 0   FOREIGN 1
after   /nix/store/r966ks7a5w9dp99c7vi2332c9z7pl7x3-node_modules   956 links   DANGLING 0   FOREIGN 0
```

The single foreign link, reproduced exactly:

```
FOREIGN <out>/node_modules/.pnpm/node_modules/@jmmaloney4/sector7 -> <out>/packages/sector7
```

A set-difference of the two trees' relative symlink paths returns **exactly one line** — `./node_modules/.pnpm/node_modules/@jmmaloney4/sector7`. Nothing else was added or removed, so no dependency-store or `.bin` link was stripped as collateral. Two extra checks beyond the ADR-047 predicate: no symlink in the post-bump tree resolves outside `$out` at all, and `.pnpm/node_modules/@jmmaloney4/` is now an empty directory.

The authoring report's caveat is correct and worth repeating — the classification predicate is byte-identical to the upstream strip rule, so "0 foreign after" is partly true by construction. The load-bearing independent evidence is the regression guard, which I checked against jackpkgs source at both revs:

| | `496252f8` | `250cd0bd` |
|---|---|---|
| `dontCheckForBrokenSymlinks` | `= true` (`modules/flake-parts/nodejs.nix:116`) | absent |
| `installPhase` | inline copy (line 118) | `jackpkgsLib.nodejs.captureNodeModules` (line 156) |

So `noBrokenSymlinks` genuinely ran against the post-bump tree during the build and passed.

**Footprint note:** `pnpm-lock.yaml` has exactly two importers (`.` and `packages/sector7`). The one other `workspace:*` specifier in the repo lives in `packages/sector7/examples/stack/package.json`, which the `packages/*` workspace glob does not match — it is not an importer, so it materialises no link. That is why sector7 carried one instance of the bug rather than garden's fan-out.

### 3. The three repo-side fixes

1. **`pnpmDepsHash`** — the FOD rebuilt under the new pin (the captured tree above exists and `provider-version-stamped` builds through it). Correct.
2. **`jackpkgs.pre-commit.adr.directory`** — verified load-bearing, not merely accepted-and-ignored: the generated pre-commit config at this HEAD contains `adr-conflict-check --adr-dir docs/internal/designs`. The value also matches this repo's own pre-existing convention — `_dogfood-adr-management.yml` already passes `adr_glob: 'docs/internal/designs/*.md'`. ADR numbering audited directly: 35 files, `000`–`034`, no duplicates, no gaps.
3. **mdformat `$` escaping** — verified content-neutral. Normalising `\$`→`$` and stripping whitespace makes all three files byte-identical to their `origin/main` versions, so the change is escaping plus reflow only.

I also checked the blast radius of touching three ADR files: `adr-management.yml` gates placeholder creation on `--diff-filter=A` (added only), so modifying existing ADRs cannot overwrite them with stubs. No data-loss path.

### 4. Validation

`nix flake check -L` on x86_64-linux — **exit 0**, all 7 checks green (`vitest`, `tsc`, `treefmt`, `pre-commit`, `renovate-config`, `provider-version-stamped`, `provider-go-test`). `nix develop` enters cleanly. Branch is 1 ahead / 0 behind `origin/main`; no rebase needed.

Release machinery is exercised rather than merely untouched: `provider-version-stamped` builds `pulumi-resource-sector7` and asserts the binary's self-reported schema version, and it is green — so the 166-commit jump did not break the artifact other repos consume.

### 5. Two pre-existing CI observations (non-blocking, out of scope here)

- **All three self-hosted runners are `offline`.** `📌 add to project` and `📝 adr` have therefore been `queued` since the PR opened and will not complete. `mergeStateStatus=UNSTABLE` reflects that, not a failure. Combined with `❄️ nix (dogfood)` being `disabled_manually` and `📦 pnpm (dogfood)` not triggering on `pull_request`, the local `nix flake check` above is the *only* gate this change has.
- **`_dogfood-claude-review.yml` passes `ref: ${{ github.ref }}`.** On an `issue_comment` event that resolves to `refs/heads/main`, not the PR head — so the on-demand `claude review` bot would review `main` rather than the PR even with runners online. This is why no bot review appears below. Both are pre-existing conditions unrelated to this bump, and both live in shared CI machinery other repos consume, so I did not touch them.

### Review-gate note

This repo has no PR-Agent / Qodo integration (no `pr-agent`/`qodo`/`codium` reference anywhere under `.github/`), so the SOP's `/review` + `/improve` convergence loop has nothing to run against. Per the SOP's fallback, the gate here is this review plus the local validation above.

**No blocking findings. Merge-ready — not merging; awaiting the owner's go-ahead.**

jmmaloney4:
## Independent review (stage 3) — verified; two corrections, no blocking findings

Re-derived from scratch at `07c6aa4` in a fresh checkout. I did not take any number from the authoring report or from the earlier stage-2 comment. Two things in the existing record turned out to be wrong; both are corrected below.

### 1. Correction: this repo **does** have PR-Agent, and the gate had not run

The stage-2 comment concluded *"This repo has no PR-Agent / Qodo integration … so the SOP's `/review` + `/improve` convergence loop has nothing to run against."* That is incorrect. It was inferred from `grep`ping `.github/` — but PR-Agent here is a **GitHub App**, not a workflow file, so it leaves no trace in the tree.

Empirically, `pr-agent-room-of-requirement[bot]` has reviewed recent sector7 PRs (#362 ×3, #363 ×3, #364 ×4, #365 ×7 comments), posting exactly the SOP's artifacts — `## PR Reviewer Guide 🔍` with `#### (Review updated until commit …)` and `## PR Code Suggestions ✨` with `<!-- <short-sha> -->` markers.

The real reason it was silent here: **#372 is a draft**, and PR-Agent does not auto-review drafts. So a mandatory SOP Phase 5 gate had been skipped, not correctly waived. I triggered it manually. Both gates came back clean at the current HEAD:

- `/improve` → `No code suggestions found for the PR.`
- `/review` → *No major issues detected*, *No security concerns identified*, no recommended focus areas.

HEAD is unchanged at `07c6aa4` and no commits were pushed after the trigger, so this satisfies the SOP's clean-run convergence conditions. **The gate is now genuinely converged rather than assumed inapplicable.**

### 2. Correction: the `pnpmDepsHash` change has three causes, not one

The PR body attributes it solely to *"the new pin ships pnpm 11.11.0 (was 11.1.1)"*. The hash is **correct** — I re-derived it from a real fetch rather than trusting it (set the hash to the all-`A` placeholder and forced the FOD to build):

```
hash mismatch in fixed-output derivation '…-pnpm-pnpm-deps.drv':
         specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
            got:    sha256-LeFkdfR+2ilF+O2PnAHUY9vjAdn3JB41RIifDabMxao=
```

Exact match with the committed value, and it is the product of a **successful** fetch (a broken fetch errors out rather than reporting a `got:` hash). Resulting tree is 176 MB — not a truncated fetch. But three things in `modules/flake-parts/nodejs.nix` changed between the pins, not one:

| | `496252f8` | `250cd0bd` |
|---|---|---|
| pnpm | `pkgs.pnpm_11` (11.1.1) | `pkgs.pnpm_11` (11.11.0) |
| `nodejs-slim` | default | overridden to `nodejs-slim_latest` (NixOS/nixpkgs#525627 workaround) |
| `fetcherVersion` | **3** | **4** |

The `fetcherVersion` 3→4 bump alone changes the FOD. I have corrected the PR body. No correctness impact — just an incomplete causal claim in a section whose entire purpose is to justify that a hash change is legitimate.

### 3. ADR-047 measurement, independently reproduced (both directions)

I rebuilt the **pre-bump** devshell from `HEAD~1` rather than only inspecting the after-state, so the bug is demonstrated to have existed and to be fixed:

```
before  /nix/store/gd11jdqxx0327brb6r84abjn8qnz4akg-node_modules   957 links   DANGLING 0   FOREIGN 1
after   /nix/store/r966ks7a5w9dp99c7vi2332c9z7pl7x3-node_modules   956 links   DANGLING 0   FOREIGN 0
```

The single foreign link, reproduced verbatim:

```
FOREIGN <out>/node_modules/.pnpm/node_modules/@jmmaloney4/sector7 -> <out>/packages/sector7
```

and `<out>/packages/sector7` contains exactly one entry — `node_modules` — no `package.json`, no source. Confirmed skeleton. Post-bump, `.pnpm/node_modules/@jmmaloney4/` is an empty directory and **zero** symlinks anywhere in `$out` resolve into `packages/sector7`. Exactly one link removed; nothing legitimate stripped.

I agree with stage 2 that the predicate is partly true by construction, so the load-bearing evidence is the guard flip, which I checked against jackpkgs source at both revs: `dontCheckForBrokenSymlinks = true` at `nodejs.nix:116` under `496252f8`, absent under `250cd0bd`, where `installPhase = jackpkgsLib.nodejs.captureNodeModules` (line 156). `noBrokenSymlinks` really did run against the post-bump tree.

### 4. The other two repo-side fixes

**`adr.directory`** — verified *necessary*, not merely accepted. The option default is `docs/internal/decisions` (`modules/flake-parts/pre-commit.nix:298`); `docs/internal/decisions` has **never existed** in this repo's history (`git log --diff-filter=A` returns nothing). I deleted the line and rebuilt the check, which fails exactly as claimed:

```
adr-conflict-check: directory not found: docs/internal/decisions
```

With the line present, the generated config carries `adr-conflict-check --adr-dir docs/internal/designs`. ADR numbering audited: 35 files, 35 unique numbers, `000`–`034` — contiguous by pigeonhole, no duplicates.

**ADR reformats** — proven content-neutral mechanically: normalising `\$`→`$` and collapsing whitespace makes all three files **byte-identical** to their pre-bump versions. Rendering verified with `cmark`: `\$0.36` → `$0.36`. Escaping plus reflow only.

### 5. Lockfile scope

13 nodes moved. I walked the lock graph from `root`: **every one** is reachable only via `root.jackpkgs`, and `jackpkgs` is the sole independently-declared pin (`nixpkgs`/`flake-parts`/`systems` are all `follows`). No node's `original` (unlocked) ref changed anywhere — so no declared pin silently moved. `5b6b6f27` (ADR-047) is a verified **ancestor** of `250cd0bd`; old pin → new pin is a fast-forward of 166 commits with no history rewrite, and the 2 commits after the fix are docs-only ADR-044 status flips. One structural edge change: `pre-commit-hooks` dropped its `gitignore` follows-edge (upstream removed the input). Benign.

### 6. Hunting for breakage across the 166-commit jump

- **Check set is identical** before and after — `pre-commit`, `provider-go-test`, `provider-version-stamped`, `renovate-config`, `treefmt`, `tsc`, `vitest`. ADR-045 added/removed nothing here.
- **mypy → ty (ADR-046):** hook set went `nbstripout pytest ruff treefmt vitest` → **`+adr-conflict-check +ty`**. `pre-commit.settings.hooks.mypy.enable = lib.mkForce false` still evaluates (the option survives), so it is now a harmless no-op. Worth noting for later: `ty` is enabled and **not** force-disabled, where its predecessor `mypy` deliberately was. Today that is inert — this repo has **0 Python files** — but if Python is ever added, `ty` will enforce where `mypy` was intentionally silenced. Non-blocking; flagging so it is a decision rather than a surprise.
- **pnpm fail-closed on tarball integrity:** pnpm 11.4.0+ now fails closed on GitHub-release `.tgz` deps whose lockfile resolution lacks an `integrity` field (`ERR_PNPM_MISSING_TARBALL_INTEGRITY`), and this bump crosses that boundary (11.1.1 → 11.11.0). sector7's own lockfile has **zero** tarball resolutions, so it cannot bite here. **Downstream note:** sector7 ships as a GitHub-release tarball, so consumers bumping to this same jackpkgs may hit this on *their* sector7 lockfile entry if it lacks `integrity:`. Out of scope for this PR, but worth expecting.

### 7. Validation

`nix flake check -L` (x86_64-linux) — **exit 0**, all 7 checks green. I re-ran `vitest` with `--rebuild` rather than accepting the cached result: **278 tests / 24 files passed**, matching the claim. `provider-version-stamped` is green, so the release artifact other repos consume still builds and self-reports correctly across the 166-commit jump. Release/versioning machinery untouched.

### 8. CI coverage — disclosure is accurate

Confirmed independently. `❄️ nix (dogfood)` is `disabled_manually` (last runs 2026-05-20, both push-to-`main` runs failing). `📦 pnpm (dogfood)` triggers only on `workflow_dispatch`, push to `main`, and `v*` tags — **not** `pull_request`. `nix.yml`/`pnpm.yml` are `workflow_call`-only reusable workflows that never self-trigger. Actual runs on this head: only `claude review` (skipped) and two queued jobs. All three self-hosted runners are **offline**, so `mergeStateStatus=UNSTABLE` reflects permanently-queued jobs, not failures.

So there is genuinely **no CI gate** on this PR, the PR body says so plainly, and I did not re-enable anything.

---

**Verdict: no blocking findings.** Both corrections above are to the *record*, not to the code — the diff itself is correct as written. PR-Agent is converged clean at `07c6aa4`, local validation is green, and this is the only gate this change has.

**Merge-ready — explicitly NOT merging; awaiting the owner's go-ahead.**

jmmaloney4:
## Addendum — deeper sweep of the 166-commit range

Follow-up to my review above. I swept the full `496252f8..250cd0bd` range for consumer-visible changes rather than only confirming the reported ones. Nothing blocking; three things worth recording.

### 1. The runtime half of ADR-047, verified directly

The captured-tree classification proves the *store* side. The generated `vitest` pre-commit hook proves the *runtime* side, and it is the more convincing artifact. At this HEAD it ends with:

```bash
jackpkgs_ensure_dir node_modules/@jmmaloney4
jackpkgs_link "$(pwd)"/packages/sector7 node_modules/@jmmaloney4/sector7
```

`mkWorkspaceRuntime` resolves `@jmmaloney4/sector7` to **`$(pwd)/packages/sector7`** — the live checkout — and the store tree it links from is the cleaned `r966ks7a…` capture. So the two halves compose exactly as ADR-047 claims: the poisoned store-internal link is stripped from the capture, and the workspace link is re-established at root level pointing at live source. That is the actual `ERR_MODULE_NOT_FOUND` path, closed.

### 2. `mkWorkspaceRuntime` now destructively replaces live `node_modules` entries (by design, and it warns)

Separately in this range (jackpkgs #358/#359), `ln -sfn` was replaced with a helper that `rm -rf`s the target first:

```bash
jackpkgs_link() {
  if [ -d "$2" ] && [ ! -L "$2" ]; then
    echo "jackpkgs: replacing real directory $2 with a link into the pinned dependency tree (re-run 'pnpm install' to restore a local install)" >&2
  fi
  rm -rf -- "$2"
  ln -s -- "$1" "$2"
}
```

Practical consequence for anyone working in a sector7 checkout: the `vitest` hook is enabled by default, so a local `git commit` will replace a real `pnpm install`ed `node_modules/<pkg>` with read-only store links, after which `pnpm install` silently no-ops. Recovery is `rm -rf node_modules && find . -name node_modules -type l -delete && pnpm install`; the devshell `shellHook` in this same range also detects the state and prints it.

This is the fix side of that behaviour, not the cause — the old pin produced the same clobber *plus* poisoned links. It only replaces `node_modules` entries, never `packages/sector7` source. Recording it because it is a genuine change in what `git commit` does to a working tree, and it is easy to mistake for corruption the first time you see it. Not a blocker.

### 3. `pre-commit.settings.hooks.mypy.enable = lib.mkForce false;` is now dead config

Confirmed it still **evaluates** — `cachix/git-hooks.nix` at the newly pinned `bca82caa` still declares a `mypy` hook, so the option exists and `mkForce false` resolves cleanly. That was the specific risk and it is clear.

But jackpkgs no longer *sets* `hooks.mypy` at all; ADR-046 replaced it with `hooks.ty`, gated on `jackpkgs.checks.python.ty.enable`, which defaults to `true`. So the line now suppresses a hook nothing enables, while its successor runs unsuppressed. `ty` has `files = "\.py$"` and this repo has 0 Python files, so it never executes — it only adds `pkgs.ty` + `pkgs.python3` to the `pre-commit` check closure.

**I deliberately did not change this in this PR.** `pytest` and `ruff` were already enabled-and-tolerated with no Python before the bump, so `ty` joining them is consistent with the repo's existing posture, and swapping the line for `jackpkgs.checks.python.ty.enable = false;` is a behaviour decision that does not belong in a dependency bump. Flagging it as a follow-up: the dead line should either be removed or replaced, so a future reader is not misled into thinking a type checker is being suppressed when the hook it names no longer exists.

### 4. Checked and clear

- **`just.cut`:** `lib/justfile-helpers.nix` changed multi-line command indentation in this range. sector7's `cut` block sets only `enable`/`files`/`commitMessage` with no custom commands, so the generated justfile text is unaffected. Release machinery untouched and `provider-version-stamped` is green.
- **`adr.directory` placement:** the new `adr.*` options are declared per-system, so setting them at the flake top level would fail as an unknown option. sector7 sets it inside `perSystem` (alongside `pre-commit.settings.hooks.*`), which is correct — and empirically it flows through to `--adr-dir docs/internal/designs`.
- **Consumer nixpkgs requirements:** `fetcherVersion = 4`, `nodejs-slim_latest`, and `python3Packages.mdformat-myst` all must exist in the consumer's nixpkgs. sector7 `follows` jackpkgs/nixpkgs, so it picks up `6cdc7fc` automatically and all three resolve — confirmed by the green build rather than by inspection.
- **Working-tree hygiene:** my `nix develop` invocations left the checkout clean (`git status` empty, no `node_modules` created), so nothing in this review mutated the PR branch.

jmmaloney4:
## `/improve` triage — the suggestion was correct; fixed in `fbc0997`

PR-Agent's `/improve` (score 8/10, marker `<!-- 07c6aa4 -->`, i.e. current HEAD) flagged line 200 of `020-cloudflare-worker-uptime-monitor.md`: `($5/mo)` and `$0.75/GB-month` left bare while `\$5/million` on the same line was escaped.

**My first instinct was that this was a false positive** — I reasoned from the dollarmath spec that both candidate closing delimiters are preceded by whitespace and therefore no inline-math span could form, and that `treefmt` passing meant mdformat had already produced the canonical escaping. Both of those arguments are wrong, so I tested it instead of trusting either.

### Verification

Rendering the line through `markdown-it` + the `dollarmath` plugin (what MyST uses, and what jackpkgs' newly-added `mdformat-myst` targets):

```
Workers Paid (<span class="math inline">5/mo) for 1-minute intervals or higher
monitor counts. D1 storage at 5 GB free, then</span>0.75/GB-month. KV writes at $5/million on paid.
```

The two bare `$` pair up and swallow ~90 characters of the sentence into a math element. GitHub's own renderer also supports `$…$` inline math, so this is not a hypothetical downstream-renderer concern.

I then rendered **every** ADR under `docs/internal/designs/` whole-file (so code fences are respected — a naive line-by-line scan produces a false positive on `010:39`, where `${BASE_REF}`/`$ADR_GLOB` sit inside a fenced block):

| corpus state | inline-math spans across all 35 ADRs |
|---|---|
| before this PR (`ec9d2e6`) | **1** — this line |
| at `07c6aa4` | **1** — unchanged |
| with `fbc0997` | **0** |

So it is exactly one defect in the whole corpus, it **pre-dates this branch**, and the bump's mdformat pass escaped only the third delimiter and left the pair intact.

### Why I fixed it here rather than deferring

Normally I would keep pre-existing debt out of a dependency bump. Three things make this the exception:

1. The bump **already modifies this exact line** — it is inside the PR's diff, not adjacent to it.
2. The PR's own claim for these three files is *"escaping only, no content change"* / renders identically. That claim was not quite true while a math span was eating the sentence; now it is.
3. It is two characters, and `treefmt` is stable on the result — mdformat does not strip the added escapes (`formatted 191 files (0 changed)`), which I confirmed before committing rather than assuming.

`nix flake check -L` is green across all seven checks on `fbc0997`.

### Scope check

I deliberately did **not** widen this. The whole-file scan found no other math-mangled line anywhere in `docs/internal/designs/`, so there is no adjacent debt to sweep up, and I touched nothing outside the file the PR was already changing.

jmmaloney4:
## Merge gate — converged at `fbc0997`. Not merging.

### PR-Agent convergence

| gate | state at `fbc0997` |
|---|---|
| `/review` | `#### (Review updated until commit …/fbc09971cd653c95d00a97075bc4ba41f615ca1c)` — SHA matches HEAD exactly. No major issues, no security concerns, no focus areas. |
| `/improve` | `No code suggestions found for the PR.` |

This is a *verified current-head* review, not the SOP's clean-first-run fallback: the persistent review carries an explicit SHA anchor equal to HEAD, plus a `final_update_message` confirming the update. The one ≥-threshold suggestion raised during the loop (8/10, `020` line 200) was fixed in `fbc0997` and `/improve` came back empty on the re-run.

Worth recording for the audit trail: this repo's PR-Agent is a **GitHub App**, so it leaves no trace under `.github/`. An earlier review inferred from `grep` that no PR-Agent integration existed and treated Phase 5 as inapplicable. It does exist — it simply does not auto-review **drafts**, which is why it was silent. The gate ran only after being triggered manually, and again on ready-for-review and on push.

### Validation

`nix flake check -L` (x86_64-linux) — **exit 0**, all seven checks green at `fbc0997`: `vitest`, `tsc`, `treefmt`, `pre-commit`, `renovate-config`, `provider-version-stamped`, `provider-go-test`. `vitest` was re-executed with `--rebuild` rather than accepted from cache: **278 tests / 24 files**.

### Merge state

`mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE`. The `UNSTABLE` is not a failure: the only GitHub checks are `add-to-project` and `adr-management`, both permanently `QUEUED` because all three self-hosted runners report `status=offline`. Combined with `❄️ nix (dogfood)` being `disabled_manually` and `📦 pnpm (dogfood)` not triggering on `pull_request`, **the local flake check above is the only gate this change has** — which the PR body states plainly. I did not re-enable any workflow.

### Non-blocking, documented

- `pre-commit.settings.hooks.mypy.enable = lib.mkForce false;` is now dead config (ADR-046 replaced the hook with `ty`). It still evaluates cleanly — that was the risk, and it is clear. Left as-is deliberately; changing it is a behaviour decision, not a bump. Follow-up.
- `mkWorkspaceRuntime` now destructively replaces live `node_modules` entries with store links on `git commit`. By design (jackpkgs #358/#359), it warns, and this bump is the fix side of that behaviour rather than its cause.
- Downstream: sector7 ships as a GitHub-release tarball, and pnpm 11.4.0+ (crossed by this bump) fails closed on tarball deps whose lockfile entry lacks `integrity:`. sector7's own lockfile has zero tarball resolutions, so it is unaffected — but consumers bumping to this same jackpkgs may need `integrity:` lines added.

---

**Every gate item passes. Explicitly NOT merging** — the owner asked for review to convergence only. This PR is merge-ready and awaiting their go-ahead.

jmmaloney4:
## Merge gate — rebased onto current `main`, re-validated, converged at `b595552`

The branch had gone `diverged` (6 behind / 2 ahead). Brought current by **rebase** (as requested) and re-validated from scratch, since a rebase moves the SHA and invalidates the prior convergence at `fbc0997`.

### How it was brought current

`git rebase origin/main` — **zero conflicts**. The 6 new base commits (#373, #374, #375) touch only `packages/sector7/**` TypeScript and `provider/**` Go; they touch no `flake.nix`, `flake.lock`, `package.json`, or `pnpm-lock.yaml`. So the lockfile/flake conflict surface never materialised and `pnpmDepsHash` is **not** implicated by the rebase — it remains the value derived for the new pin (`sha256-LeFkdfR+...`).

Force-pushed with `--force-with-lease` (`fbc0997` → `b595552`). Post-rebase the branch is 0 behind / 2 ahead and the diff against `main` is still exactly the intended 5 files.

### Lock integrity

- `jackpkgs` pinned at `250cd0bd9c24db7ab600d8b3f6d39584344dc6bc`.
- `gh api compare 5b6b6f27...250cd0bd` → `behind_by: 0` (ahead_by 2) — the ADR-047 commit is contained.
- **No unrelated input drift.** sector7's only true direct input is `jackpkgs`; its `nixpkgs`, `flake-parts` and `systems` are all `follows` edges onto jackpkgs. Every other changed lock node (`nixpkgs_2`, `flake-parts_2`, `treefmt`, `pre-commit-hooks`, `fenix`, `home-manager`, `uv2nix`, …) is an input *of jackpkgs*, re-locked as a consequence of moving the pin — not an independent bump.

### Validation at `b595552`

| gate | result |
|---|---|
| `nix flake check -L` | **exit 0**, no errors. Nothing failed, so no sibling check was cancelled/left unbuilt. |
| ADR-047 symlink classification | **956 symlinks, 0 DANGLING, 0 FOREIGN** |

The symlink figure matches the expected post-fix measurement exactly (pre-fix was 957 / 1 FOREIGN; the one removed link was `node_modules/.pnpm/node_modules/@jmmaloney4/sector7 -> <out>/packages/sector7`, whose target was a skeleton containing only a nested `node_modules/`). `noBrokenSymlinks` is back on as the guard.

### Review gate at `b595552`

- `/review` persistent comment header now reads `(Review updated until commit .../b5955524f403b01ef4820c68b4498dc5ec1bee22)` — marker equals HEAD. No major issues, no security concerns.
- `/improve` → `No code suggestions found for the PR.`
- Gemini posted a daily-quota warning; per SOP it is opportunistic and explicitly non-blocking.

**Triage of the one `/review` focus area (false positive):** it flagged that `pnpmDepsHash` and `jackpkgs.pre-commit.adr.directory` change with "no `flake.lock` update visible anywhere in the diff", warning of eval failure and a hash mismatch — but it self-qualified with *"if `flake.lock` already pins `jackpkgs` at the ADR-047 commit, this can be ignored."* That precondition holds: `flake.lock` **is** in the diff and pins `250cd0bd`, proven `behind_by: 0` against `5b6b6f27` above. The bot simply did not have the (large, generated) lockfile hunk in its context. Independently disproven by the green `nix flake check -L`, which would have failed on both counts it predicted.

### CI note

`mergeable_state=unstable` here is **queueing noise, not failure**. The only two check runs at this HEAD are `add-to-project` and `adr-management / push placeholder to main`, both `queued` on offline self-hosted runners; zero checks are failing. sector7 has no nix gate on pull requests (`❄️ nix (dogfood)` is `disabled_manually`, `📦 pnpm (dogfood)` has no `pull_request` trigger), so local `nix flake check -L` is the authoritative gate — and it is green.

Merging with a merge commit (squash and rebase merges are disabled on this repo).

PR #374, opened 2026-08-13
refactor(provider): extract checkutil.RequireNonEmpty helper

jmmaloney4:
## Summary

- Extract `provider/internal/checkutil` (`RequireNonEmpty`), modeled on the `diffutil`/`httpx`/`kube` internal packages: a small, doc-commented package with its own test file.
- Replace the hand-rolled "required non-empty string" idiom in all 7 call sites with `checkutil.RequireNonEmpty`:
  - `provider/r2/object.go`
  - `provider/r2/zonecachepurge.go`
  - `provider/attic/cache.go`
  - `provider/onepassword/item.go`
  - `provider/litellm/key.go`
  - `provider/matrix/room.go`
  - `provider/matrix/botaccount.go`
- No behavior change to failure `Property` names. The only wording change is `r2.ZoneCachePurge`'s `zoneId`/`apiToken`/`trigger` messages, which moved from `"<field> is required and must be a non-empty string"` to the `"<field> is required"` phrasing every other one of the 7 files already used independently — per the issue, consolidating on the phrasing 6/7 sites had already converged on rather than preserving the one outlier.
- Non-"required non-empty string" validations (file-path checks, files/hosts array validation, preset enum, title `TrimSpace`, field-label dedup, retention-period sign check) are left untouched — they're a different idiom, not in scope.

## Test plan

- `cd provider && go build ./...` — passes
- `cd provider && go test ./...` — all packages pass, including the new `provider/internal/checkutil` package
- `cd provider && go vet ./...` — clean
- `nix fmt` — no changes (already formatted)
- New unit tests in `provider/internal/checkutil/checkutil_test.go` cover: missing fields reported in order, all-present produces no failures, no-fields-is-a-noop, and appending to a pre-populated failures slice (matching how callers pass the slice `infer.DefaultCheck` already returned).
- Existing per-resource Check tests (`r2/object_test.go`, `matrix/matrix_test.go` `TestRoomCheck`/`TestBotAccountCheckRequiresPassword`, etc.) continue to pass unmodified — they assert on failure `Property` names only, never exact `Reason` text, so the wording consolidation doesn't touch them.

## Risks

- Low risk, mechanical refactor. The only externally-observable change is the `Reason` string on 3 `ZoneCachePurge` failures (`zoneId`, `apiToken`, `trigger`); `Property` names are unchanged, and no test or consumer asserts on that exact string.
[...]

PR #373, opened 2026-08-13
chore(attic,k8s): retire port-forward.ts and mintAtticToken

jmmaloney4:
## Summary

Retires two TypeScript modules that were superseded by the Go plugin
(garden ADR 163) and no longer had any in-repo caller besides their own
test:

- `packages/sector7/k8s/port-forward.ts` — `provider/internal/kube/kube.go`'s
  doc comment says outright "this replaces `packages/sector7/k8s/port-forward.ts`."
  Its only expected importer was its own test,
  `packages/sector7/tests/port-forward.test.ts`. Both are deleted; the now-empty
  `packages/sector7/k8s/` directory drops out with them.
- `packages/sector7/attic/token.ts`'s `mintAtticToken` — `provider/attic/token.go`'s
  `MintToken` hand-encodes the same non-obvious Attic permission-claim wire
  format. `mintAtticToken`, its private helpers (`permissionClaim`, `base64url`),
  and the now-unused `ATTIC_CLAIM_NAMESPACE` / `MintAtticTokenArgs` are removed
  from `attic/token.ts`, and the `mintAtticToken` export is dropped from
  `attic/index.ts`. `parseDurationSeconds` (and the still-shared
  `AtticCacheGrants` / `AtticCachePermissionFlags` types, which `config-types.ts`
  still imports) are kept — `attic/admin.ts` still needs `parseDurationSeconds`.

Verified dead in-repo before deleting: repo-wide grep found no importer of
`k8s/port-forward` or user of `mintAtticToken` outside each module's own test
file (plus one historical mention in `docs/internal/designs/032-attic-cache-token-resources.md`,
which is a design-doc reference, not a code dependency). Neither symbol is
exposed via a package.json `exports` subpath of its own (`port-forward.ts` was
never publicly exported at all; `mintAtticToken` was exported only from the
`./attic` subpath barrel).

## Breaking change

`mintAtticToken` is a removed public export from the `@jmmaloney4/sector7/attic`
subpath. This is unlikely to affect anyone in practice — the Go plugin already
replaced its behavior for every in-repo consumer (`AtticCache`/`AtticToken` now
retype onto `provider/attic`, garden ADR 163) — but flagging explicitly per the
issue: any external consumer that imported `mintAtticToken` directly must switch
to the plugin-backed `AtticToken`/`AtticCache` resources, or vendor the JWT-minting
logic themselves.

## Test plan

- [x] Repo-wide grep for `port-forward` imports and `mintAtticToken` usage —
[...]

PR #375, opened 2026-08-13
feat(r2): retype ZoneCachePurge onto the resource plugin

### LINKED ISSUE #366: r2: finish ZoneCachePurge migration to the Go provider plugin
## Summary

`ZoneCachePurge` was ported to the Go provider plugin (`provider/r2/zonecachepurge.go`, registered as `sector7:r2:ZoneCachePurge` in `provider/provider.go:98`) and picked up a real bugfix along the way: the old TS implementation truncates by UTF-16 code unit, the Go port truncates by rune. But the TS surface (`packages/sector7/r2/r2object.ts:440-678`) still constructs a `pulumi-nodejs:dynamic:Resource` with its own full `check`/`diff`/`create`/`update`/`delete`, never the Go module. Every caller today goes through the old TS path; the fixed Go implementation is registered and deployed but unreachable.

## Why it matters

The same purge/validation logic is maintained twice, and the fix that motivated the Go port never reaches production callers.

## Suggested fix

Retype `ZoneCachePurge` onto the plugin the same way `R2Object` and `D1Query` already were: a thin TS `CustomResource` wrapper with a URN alias for backward compatibility, deferring all logic to `provider/r2/zonecachepurge.go`. Delete the duplicate CRUD/validation in `r2object.ts` once the alias is in place.

## Context

Surfaced by an architecture-deepening review of `packages/sector7/` + `provider/`.

jmmaloney4:
## Summary

`ZoneCachePurge` was already ported to the Go provider plugin (`provider/r2/zonecachepurge.go`, registered as `sector7:r2:ZoneCachePurge` in `provider/provider.go:98`) — including a real bugfix over the old TS implementation (resource-id truncation of `trigger` is by rune, not UTF-16 code unit) — but the TS surface in `packages/sector7/r2/r2object.ts` still constructed a `pulumi-nodejs:dynamic:Resource` with its own full `check`/`diff`/`create`/`update`/`delete`, so the Go implementation was unreachable from every real caller.

This PR retypes `ZoneCachePurge` onto the plugin the same way `R2Object` (#364) and Attic `Cache`/`Token` (#365) already were: a thin `CustomResource` wrapper aliased to the old dynamic-resource URN, so the cutover is a no-op `pulumi preview` (engine rewrites the URN in place, no purge API call during the switch) and existing deployed instances aren't force-replaced.

## What changed

- `packages/sector7/r2/r2object.ts`: `ZoneCachePurge` now `extends pulumi.CustomResource`, registers under `sector7:r2:ZoneCachePurge`, aliases `{ type: "pulumi-nodejs:dynamic:Resource" }`, and marks `apiToken` as a Pulumi secret on the input side. Deleted entirely: `zoneCachePurgeProvider` (the dynamic-provider `check`/`diff`/`create`/`update`/`delete` implementation), `purgeZoneCacheApi` (native `fetch()`-based Cloudflare Purge Cache call), the `ZoneCachePurgeInputs` interface, and `rejectCloudProviderOptions` (now dead — nothing calls it once `ZoneCachePurge` accepts `provider`/`providers` like every other plugin-backed resource). `purgeZoneCache`'s `opts` parameter is now `CustomResourceOptions` instead of the dynamic-resource-only `DynamicResourceOptions`.
- `provider/r2/zonecachepurge.go`: untouched — already correct and registered per the issue.
- `packages/sector7/tests/zone-cache-purge.test.ts`: deleted. It exercised the dynamic provider's own `check`/`diff`/`create`/`update`/`delete` logic directly; that coverage now lives in `provider/r2/zonecachepurge_test.go` (already existed before this PR) and is superseded, not lost.
- `packages/sector7/tests/r2object.test.ts`: added a `purgeZoneCache` construction block mirroring the existing `R2Object construction` block — asserts the plugin token + alias, that `provider`/`providers` are now accepted (the old dynamic resource threw on them), that `apiToken` is wrapped in `pulumi.secret`, and that `dependsOn` from `args` still merges into resource options. Also removed the now-unused `dynamic.Resource` mock scaffinding from the `@pulumi/pulumi` mock (nothing in `r2object.ts` uses `dynamic` anymore).
- `packages/sector7/README.md`: updated the `r2` subpath row to say "backed by the sector7 resource plugin" instead of "dynamic resources", matching the `d1` row's wording.

## Test plan

```
$ cd provider && go build ./... && go test ./...
ok  	.../provider  ...
ok  	.../provider/r2  ...
(all packages pass, unchanged)

$ pnpm install --frozen-lockfile
$ pnpm --filter './packages/*' -r --if-present run test
 Test Files  23 passed (23)
      Tests  242 passed (242)

$ pnpm exec tsc --noEmit
(clean)

$ nix build .#checks.x86_64-linux.provider-go-test .#checks.x86_64-linux.vitest \
    .#checks.x86_64-linux.tsc .#checks.x86_64-linux.treefmt --no-link -L
(all four green)
```

## Risks

- Low. This mirrors an already-twice-validated pattern (#364, #365). The URN alias is what makes the cutover safe for any stack with a deployed `ZoneCachePurge` instance — the engine matches existing dynamic-resource state and rewrites the URN in place rather than planning a delete+create, so `pulumi preview` on the next deploy should show no changes for this resource (beyond picking up the plugin-backed implementation going forward).
- The Go implementation fixes a real (if narrow) bug: `trigger.slice(0, 12)` in the old JS counted UTF-16 code units, which could split a multibyte character; the Go port truncates by rune. Any caller whose `trigger` happens to contain multibyte characters at exactly that boundary will see a different (but valid, not corrupted) resource id on first apply after this lands — this is an id format change, not a behavior regression, and `zoneId`+`trigger` together already form the purge identity.

Closes #366


jmmaloney4:
## Merge-gate audit

- `cd provider && go build ./... && go test ./...`: pass, unchanged (Go side untouched — already correct per the issue).
- `pnpm install --frozen-lockfile` + `pnpm --filter './packages/*' -r --if-present run test`: 23 test files / 242 tests, all passing.
- `pnpm exec tsc --noEmit`: clean.
- `nix build .#checks.x86_64-linux.{provider-go-test,vitest,tsc,treefmt} -L`: all four green.
- Mirrors the already-twice-validated retype pattern from #364 and #365; URN alias to the old `pulumi-nodejs:dynamic:Resource` type token means existing deployed `ZoneCachePurge` instances migrate via a no-op `pulumi preview`, not a force-replace.
- PR-Agent `/review`: no security concerns, no major issues detected. `/improve`: no code suggestions found. Both current as of HEAD `31a8a87`.
- `gh pr view --json mergeable,mergeStateStatus`: `MERGEABLE` / `UNSTABLE` — only outstanding checks are `add-to-project` and the `adr-management` placeholder push, queued on self-hosted runners (same non-blocking pattern as #373, #374, #365, #364).

Ready to merge (`--merge`).

